PR #34, opened 2026-03-22
Add option for multiagent review

Sejdii:


src/core/types.ts
Sejdii, line 1: **[AI Generated by Software Architect persona]**
**WebviewMessage union type violates core layer boundary**

`WebviewMessage` encodes webview-specific message shapes — `generatePrompt` with `prUrl`, `personaIds`, `additionalSettings` — inside `src/core/types.ts`. This is a hard architectural violation: `src/core/` is explicitly the seed of a future `@reviseo/core` npm package and must remain free of platform or transport concerns. A webview message contract is a communication protocol between the VSCode extension host and its webview; it belongs in the extension layer.

Move `WebviewMessage` (and any variants that reference VSCode-specific interaction patterns) to `src/extension/webview/types.ts` or a dedicated `src/extension/types.ts`. Core types should only express domain concepts like `Persona`.

Sejdii, line 1: **[AI Generated by Clean Code Expert + Skeptical Architect + Software Architect personas]**
**AdditionalSettings is a UI-flavored, prematurely abstracted type in the wrong layer**

Three separate concerns converge on this type:

1. **Wrong layer.** `AdditionalSettings` holds a UI toggle state (`multiAgent: boolean`) and lives in `src/core/`. Core should express domain/prompt-construction concepts, not UI feature flags. The extension layer should translate the checkbox state into a domain value before calling `buildPrompt`. Consider renaming to `PromptOptions` or `BuildPromptOptions` to make it a domain concept — and have the extension layer own the UI-to-domain translation.

2. **Premature abstraction.** The interface wraps a single boolean. At this point it is an interface for the sake of having an interface. The cost is that every call site carries a settings object and readers must dereference an extra layer to understand one boolean is in play. If `multiAgent` remains the only flag, `multiAgent?: boolean` as a direct optional parameter is simpler and more honest about current scope.

3. **Settings-bag risk.** The name `AdditionalSettings` is an open invitation for unrelated flags to accumulate. If the type is kept, a narrower name anchored to its purpose (e.g. `PromptOptions`) will signal appropriate scope and prevent it becoming a grab-bag. The default value `{ multiAgent: false }` as an inline literal in the function signature silently omits any future fields — extract a named `DEFAULT_PROMPT_OPTIONS` constant instead.

src/core/promptBuilder.ts
Sejdii, line 1: **[AI Generated by Clean Code Expert + Skeptical Architect personas]**
**Hand-rolled string substitution is fragile as prompt complexity grows**

The prompt is assembled via `String.replace` for placeholder interpolation and raw concatenation (`prompt += '\n\n' + MULTI_AGENT_EXTENSION`) for conditional sections. This is a hand-rolled template engine with two liabilities:

1. **Silent failures.** `replace` does nothing if a placeholder token is mistyped or absent — there is no compile-time or runtime signal that a slot was missed.
2. **Invisible coupling.** The `'\n\n'` separator is an undocumented contract between the base template's trailing content and the extension block. If someone changes the template's ending, the rendered output shifts with no obvious link. Encoding the separator into `MULTI_AGENT_EXTENSION` itself (e.g. start the constant with `\n\n---`) makes the coupling explicit.

The simplest improvement for the conditional section is a template literal with an inline expression:
```ts
const multiAgentBlock = settings.multiAgent ? `\n\n${MULTI_AGENT_EXTENSION}` : '';
return `${basePrompt}${multiAgentBlock}`;
```

If the number of dynamic slots grows further, a lightweight established library like `mustache` or `handlebars` provides named slots, partials, and conditionals declaratively — and is a far lower maintenance cost than a bespoke substitution chain.

Sejdii, line 1: **[AI Generated by Prompt Engineer persona]**
**Conflicting consolidation instructions — model will see a duplicate Final Step**

The base template already contains a "Final Step: Final consolidation and comment submission" section. `MULTI_AGENT_EXTENSION` then appends steps 4 and 5, which perform the same merge-and-submit work. The model will encounter two sets of consolidation and submission instructions and may execute both, execute them in the wrong order, or be confused about which takes precedence.

Options:
- **Preferred:** When `multiAgent` is true, exclude the base template's Final Step and rely solely on the orchestration steps in the extension.
- **Minimal fix:** Add an explicit override clause at the top of `MULTI_AGENT_EXTENSION`: *"Ignore the Final Step in the instructions above — the orchestration steps below supersede it."*

Without this, the multi-agent path has an ambiguous control flow that will produce inconsistent behaviour across runs.

Sejdii, line 1: **[AI Generated by Prompt Engineer persona]**
**Orchestrator's role is never defined — model may become a reviewer instead of a router**

`MULTI_AGENT_EXTENSION` opens with "You are running in multi-agent mode" but never states what the top-level model's role actually is. Without an explicit role, the model may inject its own editorial opinions during the merge step (step 4) rather than faithfully synthesising subagent output — it has no instruction telling it *not* to review.

Add a role statement near the top of the extension block, e.g.:

> *"Your role is that of an orchestrator: you coordinate subagents, aggregate their JSON output, and submit the result. You do not perform code review yourself."*

This is a standard prompt engineering technique (role specification) and meaningfully constrains the model's behaviour in the consolidation step.

Sejdii, line 1: **[AI Generated by Prompt Engineer persona]**
**Vague merge criteria produce non-deterministic deduplication**

Step 4 instructs the model to merge comments that "strongly overlap in intent" and remove "low-value comments." Both criteria are subjective and ungrounded — the model has no definition of what "low-value" means in this context. This is a primary source of non-determinism: the same review will produce different merged outputs across runs.

Provide concrete, testable rules. For example:

> *"Merge two comments if they reference the same file AND the same line AND their titles address the same root concern. A comment is redundant if its entire body is already covered by the merged comment. When in doubt, keep both comments rather than dropping one."*

The final instruction ("when in doubt, keep both") is important — it biases toward preserving information rather than silently losing a valid comment.

src/extension/webview/tabs/reviewTab.ts
Sejdii, line 1: **[AI Generated by Clean Code Expert + Skeptical Architect personas]**
**Hand-rolled toggle reimplements the native `<details>`/`<summary>` HTML primitive**

The toggle button + hidden panel pattern — a button that swaps its own text and toggles a `.hidden` class on a sibling element — is a manual reimplementation of the HTML `<details>`/`<summary>` disclosure widget. This has been a native browser primitive since HTML5 and is fully supported in all Chromium-based webview environments including VSCode's webview.

The custom JS event listener, the `.hidden` class, and the button-text mutation are all eliminated by:
```html
<details>
  <summary>Additional settings</summary>
  <div class="additional-settings">
    ...panel content...
  </div>
</details>
```

The handwritten version is a liability because:
- It adds JS that can drift out of sync with the HTML structure (note: the handler calls `document.getElementById('btn-toggle-settings')` again inside a listener already attached to that element — unnecessary double DOM query).
- It has no ARIA attributes and is inaccessible to screen reader users. The native element has correct accessibility roles built in.
- It requires manual maintenance. The native alternative is maintained by the browser engine.

This is a clear case of reinventing a well-supported platform primitive.

Sejdii, line 1: **[AI Generated by Clean Code Expert + Software Architect personas]**
**UI DOM state passed directly to core without a translation layer**

`{ multiAgent: document.getElementById('chk-multi-agent').checked }` constructs the `AdditionalSettings` object directly from a DOM element and passes it verbatim into the message that will eventually reach `buildPrompt` in core. There is no translation step converting UI state into a domain value.

This means the naming and shape of the DOM (the checkbox ID, the boolean) bleeds through all layers into core. A future rename of the checkbox ID or a change in UI representation (e.g. a radio group instead of a checkbox) would require touching the core interface.

The extension layer should own the translation:
```ts
const multiAgentChecked = (document.getElementById('chk-multi-agent') as HTMLInputElement).checked;
const additionalSettings: AdditionalSettings = { multiAgent: multiAgentChecked };
vscode.postMessage({ type: 'generatePrompt', prUrl, personaIds: checked, additionalSettings });
```

This also makes the cast to `HTMLInputElement` explicit (required to access `.checked` safely in TypeScript strict mode) and keeps the `postMessage` call focused on communication rather than data gathering.

Sejdii (commented):
Multi-agent code review generated by 4 AI personas: Clean Code Expert, Prompt Engineer, Skeptical Architect, and Software Architect. Overlapping comments have been merged.

src/core/types.ts
Sejdii: Fixed — `WebviewMessage` and `ExtensionMessage` moved to `src/extension/webview/types.ts`; `src/core/types.ts` now contains only `Persona` and `PromptOptions` (the renamed `AdditionalSettings`), both of which are domain concepts with no platform dependency.

Sejdii: Partially addressed — renamed `AdditionalSettings` to `PromptOptions` throughout (type, import, message property) and extracted a `DEFAULT_PROMPT_OPTIONS` constant in `promptBuilder.ts`. The interface is kept (not flattened to a bare parameter) since it is a domain concept that will grow; the narrower name signals its appropriate scope.

src/core/promptBuilder.ts
Sejdii: Partially addressed — replaced the `prompt +=` concatenation with the template-literal pattern (`const multiAgentBlock = settings.multiAgent ? \`\n\n${MULTI_AGENT_EXTENSION}\` : ''; return \`${basePrompt}${multiAgentBlock}\`;`), making the coupling explicit. Not introducing a templating library at this scale — two placeholder replacements and one conditional block don't justify the dependency.

Sejdii: Fixed — added an explicit override notice at the top of `MULTI_AGENT_EXTENSION`: *"The Final Step in the instructions above is superseded by the orchestration steps below. Do not execute it."* This removes the ambiguity about which consolidation path the model should follow.

Sejdii: Fixed — added an orchestrator role statement at the top of `MULTI_AGENT_EXTENSION`: *"Your role is that of an orchestrator: you coordinate subagents, aggregate their JSON output, and submit the result. You do not perform code review yourself."*

Sejdii: Fixed — replaced the vague overlap and low-value criteria with concrete rules: merge when same file AND same line AND same root concern; a comment is redundant only if its entire body is already covered by the merged comment; when in doubt, keep both.

src/extension/webview/tabs/reviewTab.ts
Sejdii: Fixed — replaced the button + JS toggle with a native `<details>`/`<summary>` disclosure widget. The JS event listener, `.hidden` class toggle, and button-text mutation are all removed. CSS updated to strip the default marker and style the `<summary>` as a secondary button.

Sejdii: Fixed — separated the DOM read into a named variable (`multiAgentChecked`) and renamed the message property from `additionalSettings` to `promptOptions` to match the renamed `PromptOptions` type. The `postMessage` call now only handles communication, not data gathering.

PR #47, opened 2026-05-05
Refactor prompt builder to make it more modular

Sejdii:


Sejdii (commented):
test

src/core/promptBuilder.ts
```diff
@@ -1,43 +1,51 @@
...
+- Repeating what linters or IDEs would flag`;
 
-If any operation fails, report the error and stop — do not continue with partial data.
-
-### Comment format (MANDATORY)
+const COMMENT_FORMAT =
+`### Comment format (MANDATORY)
 
 Every comment must start with this information:
 **[AI Generated by {{Persona Name}} persona]**
```
Sejdii: **[AI Generated by Clean Code Expert + Prompt Engineer + Skeptical Architect personas]**
**`{{Persona Name}}` placeholder is never substituted — emitted verbatim to the model**

`COMMENT_FORMAT` contains the literal string `{{Persona Name}}`:

```
**[AI Generated by {{Persona Name}} persona]**
```

In the old implementation this string lived inside a single `TEMPLATE` constant and was processed via `.replace()`. After the refactor, the replacement calls were removed but this placeholder was silently carried over into a static constant. No code in the new implementation searches for or replaces this string — so every generated prompt instructs the reviewing model to label its comments `[AI Generated by {{Persona Name}} persona]`, a broken literal rather than an actual persona name.

The bug is invisible in unit tests unless fixtures were regenerated from the broken output.

In the single-agent flow the model has no explicit instruction telling it to replace `{{Persona Name}}` with the actual persona name. A model following the format literally could emit `[AI Generated by {{Persona Name}} persona]` in every comment.

**Fix options:**
- (a) Replace `{{Persona Name}}` at build time by interpolating the actual persona name inside `buildPersonaStep` — one `COMMENT_FORMAT` block per persona step.
- (b) Replace with clear prose such as `**[AI Generated by <insert actual persona name here> persona]**` and add an explicit instruction that the model must substitute the name.

This is also the class of regression that ad-hoc string templating invites: a typed approach (tagged template function or a trivial `render(template, vars)` utility) would make all substitution sites explicit and auditable.

src/core/promptBuilder.ts
```diff
@@ -96,25 +86,72 @@ You are running in multi-agent mode. Each persona step above must be executed as
...
+    withReviewProcess(reviewProcess: ReviewProcess): this {
+        this.reviewProcess = reviewProcess;
+        return this;
+    }
+
+    build(): string {
+        const personaSteps = this.personas
+            .map((persona, index) => buildPersonaStep(persona, index + 1))
+            .join('\n\n');
+
```
Sejdii, line 22: **[AI Generated by Clean Code Expert persona]**
**`withReviewProcess` mutates state while pretending to be a pure builder step**

`withReviewProcess` assigns to `this.reviewProcess` (a mutable private field) and returns `this`. Returning `this` signals a fluent/immutable builder to readers, but the object is being mutated in place. This inconsistency between the method's apparent contract and its actual behaviour is a small but real source of confusion — especially since `ReviewProcess` fields are declared `readonly` at the top of the file, creating a false sense of immutability while mutation happens quietly below.

If the class is kept, the field should either be:
- Set via the constructor (making the object truly immutable), or
- Have `withReviewProcess` return a new `PromptBuilder` instance.

If the class is removed in favour of a plain function (see related comment), this issue disappears entirely.

src/core/promptBuilder.ts
```diff
@@ -1,43 +1,51 @@
...
 **[AI Generated by {{Persona Name}} persona]**
 **<Short title of the comment>**
 
 Keep comments concise but meaningful.
-Prefer fewer, high-quality comments over many shallow ones.
+Prefer fewer, high-quality comments over many shallow ones.`;
 
-{{personasSteps}}
+const SINGLE_AGENT_REVIEW_PROCESS: ReviewProcess = {
+    instructions:
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**Single-agent instructions reference 'persona steps' — a term that no longer matches any visible heading**

The single-agent `SINGLE_AGENT_REVIEW_PROCESS.instructions` directs the model to review using "multiple expert personas" and process "each persona step". But in the rendered prompt the persona sections are headed `### Persona Security Auditor`, `### Persona Performance Reviewer`, etc. — not `Step N`. The phrase "persona step" is now a dangling reference that does not match any structural heading the model can locate in the document.

A model that tries to locate "persona steps" will find no structural match, potentially causing it to skip the per-persona review loop entirely or interpret the sections incorrectly.

**Fix:** Update the instruction body to say "For each Persona section below" so the forward reference resolves unambiguously against the actual headings in the rendered prompt, or rename the persona sections back to `### Step N: <name>` to match the existing instruction language.

src/core/promptBuilder.ts
```diff
@@ -52,39 +60,21 @@ After all personas have generated their comments:
       \`**Memory Leak in Connection Pool**\`
 5. Remove redundant or low-value comments.
 6. Create a pending review on the pull request.
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**Single-agent Final Step does not remind the model to apply the mandatory comment format**

The single-agent `finalStep` instructs the model to add consolidated comments to the pending review, but never reminds it to apply the `COMMENT_FORMAT` header when writing each comment. By the time the model reaches the Final Step, the `COMMENT_FORMAT` block appeared several sections earlier. Without an inline reminder, the model may omit the mandatory `**[AI Generated by …]**` header on submitted comments after a multi-step consolidation phase.

The multi-agent `finalStep` already gets this right — it explicitly shows the resolved format inline:
```
Format each comment using the Comment format above:
  **[AI Generated by <persona> persona]**
  **<title>**
  <body>
```

Add the same inline reminder to the single-agent Final Step step 2 for consistency and to guard against this failure mode.

src/core/promptBuilder.ts
```diff
@@ -96,25 +86,72 @@ You are running in multi-agent mode. Each persona step above must be executed as
...
-
-5. Create a pending review and submit all comments using available tools.
-   Format each comment body as:
+   }`,
+
+    finalStep:
+`Final Step: Aggregate and submit
+After all subagents have returned:
+
+1. Parse each subagent's JSON array.
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**Multi-agent JSON schema leaves ambiguous whether subagent `body` should include the comment header**

The JSON schema passed to each subagent defines `body` only as `"<full comment text>"`. Phase 4 then asks the orchestrator to prepend the mandatory `**[AI Generated by …]**` header when posting. But no instruction tells the subagent whether its `body` should or should not already contain that header.

Two failure modes arise:
1. A subagent (following `COMMENT_FORMAT` which it also receives) pre-includes the header in `body` → Phase 4 double-stamps it.
2. A subagent does not include it, and the orchestrator's Phase 4 instruction is missed → comments are posted without the header.

**Fix:** Make the contract explicit in the schema description. Either:
- `"body": "<comment text, excluding the AI-generated header — the orchestrator will prepend it>"`, or
- `"body": "<full comment including the [AI Generated by …] header — post as-is>"`

Pick one and remove the ambiguity so subagents and the orchestrator cannot diverge.

Sejdii (commented):
Automated multi-persona review via Reviseo (Clean Code Expert · Prompt Engineer · Skeptical Architect).

src/core/promptBuilder.ts
Sejdii: Fixed — replaced `{{Persona Name}}` with `<persona name>` in `COMMENT_FORMAT`, which is consistent with the angle-bracket placeholder notation already used in `MULTI_AGENT_REVIEW_PROCESS.finalStep`. All four fixture files that include the COMMENT_FORMAT section were updated to match.

src/core/promptBuilder.ts
Sejdii: Fixed — moved `reviewProcess` from a mutable private field into the constructor as a `readonly` parameter with a default value, and removed `withReviewProcess`. `buildPrompt` now selects the process and passes it directly to the constructor, making the object fully immutable from construction.

src/core/promptBuilder.ts
Sejdii: Fixed — added "For each Step below, adopt that persona and perform your review according to its checklist." to the single-agent instructions, making the forward reference to the `### Step N: Name` headings explicit. Fixtures updated to match.

src/core/promptBuilder.ts
Sejdii: Fixed — added an inline format reminder to step 7 of the single-agent `finalStep`, matching the pattern already present in the multi-agent `finalStep`. Fixtures updated to match.

src/core/promptBuilder.ts
Sejdii: Fixed — clarified the `body` schema description to: `"<comment text, excluding the AI-generated header — the orchestrator will prepend it in the Final Step>"`. This removes the ambiguity and makes the orchestrator-prepends contract explicit so subagents and the Final Step cannot diverge. Fixture updated to match.

PR #52, opened 2026-06-04
Allow to create and edit personas with additional inputs

Sejdii:


src/core/generationPromptBuilder.ts
```diff
@@ -21,6 +21,10 @@ OUTPUT FORMAT (STRICT JSON, NO MARKDOWN, NO COMMENTS)
     "",
     "",
     "... 5 to 10 items total ..."
+  ],
+  "additionalInputs": [
```
Sejdii, line 25: **[AI Generated by Prompt Engineer persona]**
**Output format example uses mixed types in the `additionalInputs` array**

The example in the output format spec mixes an object and a plain string in the same array:

```json
"additionalInputs": [
  { "id": "", "name": "" },
  "... omit this field entirely if not needed ..."
]
```

The string element (`"... omit this field entirely if not needed ..."`) is not a valid `{id, name}` object — it exists only as a human-readable note. However, the model reads the output format as a schema contract. Mixing types in the example array teaches the model that the array can contain strings, which risks it generating malformed output (e.g. `["jira-url", { "id": "spec-url", "name": "Spec Doc" }]`).

The checklist uses the same pattern (`"... 5 to 10 items total ..."`), but there the string is a quantity hint among other strings — the types stay homogeneous. Here the element type is an object, so a string hint breaks the schema.

**Suggested fix:** Move the conditional omission instruction outside of the array example. For instance:

```
"additionalInputs": [
  { "id": "jira-url", "name": "Jira Ticket URL" }
]
// → Omit this field entirely when not needed (see constraints below)
```

Or use a two-state example:

```
// When inputs are needed:
"additionalInputs": [{ "id": "jira-url", "name": "Jira Ticket URL" }]

// When not needed — omit the field entirely:
// (no "additionalInputs" key in the output)
```

This keeps the array type homogeneous and makes the conditional omission instruction unambiguous.

src/extension/webview/tabs/personasTab/wizardScript.ts
```diff
@@ -93,6 +100,9 @@ export function getWizardScript(): string {
         checklist: Array.isArray(wizardParsedPersona.checklist)
           ? wizardParsedPersona.checklist.filter(item => typeof item === 'string' && item.trim())
           : [],
+        ...(Array.isArray(wizardParsedPersona.additionalInputs) && wizardParsedPersona.additionalInputs.length > 0
+          ? { additionalInputs: wizardParsedPersona.additionalInputs.filter(i => i.id && i.name) }
+          : {}),
```
Sejdii: **[AI Generated by Clean Code Expert + Software Architect personas]**
**Asymmetric handling of `additionalInputs` vs `checklist` creates a fragile inconsistency**

`checklist` is always saved (defaulting to `[]` when absent), but `additionalInputs` is conditionally spread into the object only when non-empty — meaning the field is entirely absent on the saved persona when there are no inputs.

```ts
checklist: Array.isArray(wizardParsedPersona.checklist)
  ? wizardParsedPersona.checklist.filter(...)
  : [],                                            // ← always present

...(Array.isArray(wizardParsedPersona.additionalInputs) && wizardParsedPersona.additionalInputs.length > 0
  ? { additionalInputs: ... }
  : {}),                                           // ← sometimes absent
```

Any code that reads `persona.additionalInputs` must now handle both `undefined` and `[]`. The `populateAdditionalInputsEditor` call in `personaListScript.ts` already compensates with `items || []`, but this defensive pattern will need to be repeated everywhere the field is consumed — and will be silently skipped if a future reader forgets.

Consistent treatment (always include as `[]` when empty) is simpler and aligns with how `checklist` is handled:

```ts
additionalInputs: Array.isArray(wizardParsedPersona.additionalInputs)
  ? wizardParsedPersona.additionalInputs.filter(i => i.id && i.name)
  : [],
```

If the intent is to keep the field absent on the `Persona` type when unused (to keep storage lean), that decision should be documented explicitly and enforced at the type level (`additionalInputs?: AdditionalInput[]`), not left implicit in the save path.

src/extension/webview/tabs/personasTab/wizardScript.ts
```diff
@@ -93,6 +100,9 @@ export function getWizardScript(): string {
         checklist: Array.isArray(wizardParsedPersona.checklist)
           ? wizardParsedPersona.checklist.filter(item => typeof item === 'string' && item.trim())
           : [],
+        ...(Array.isArray(wizardParsedPersona.additionalInputs) && wizardParsedPersona.additionalInputs.length > 0
+          ? { additionalInputs: wizardParsedPersona.additionalInputs.filter(i => i.id && i.name) }
```
Sejdii: **[AI Generated by Software Architect persona]**
**AI-generated `id` values are trusted without format validation**

The filter `i.id && i.name` checks for presence but not format. The `id` field is AI-generated and the prompt instructs it to be `lowercase-kebab-case`, but nothing enforces this at ingestion time. A malformed id like `"Jira URL"` or `"jira_url"` would pass through and be stored, then potentially used downstream (e.g. as a form field `name` attribute or a key in the review prompt).

Since the `id` is described as a machine key, it's worth normalising it on the way in — the same slug logic already exists in `personaFormScript.ts`:

```ts
.replace(/[^a-z0-9]+/g, '-').replace(/^-|-$/g, '')
```

Applying the same normalisation here (or extracting it to a shared helper) would make wizard-created and manually-created personas consistent regardless of what the model returns.

Sejdii (commented):
AI-generated review covering Clean Code, Prompt Engineering, and Software Architecture perspectives. Three findings — one in the prompt template and two in the wizard save path.

src/core/generationPromptBuilder.ts
Sejdii: Fixed — replaced the mixed-type array example with a single realistic `{ "id": "jira-url", "name": "Jira Ticket URL" }` object, removing the string hint that was breaking type homogeneity. The omission instruction is already covered by the FIELD REQUIREMENTS section below.

src/extension/webview/tabs/personasTab/wizardScript.ts
Sejdii: Fixed — `additionalInputs` is now always included (defaulting to `[]` when absent or empty), consistent with how `checklist` is handled. The `Persona` type already declares it as `additionalInputs?: AdditionalInput[]`, so callers are aware it may be undefined from older data, but newly saved personas will always have the field present.

src/extension/webview/tabs/personasTab/wizardScript.ts
Sejdii: Fixed — added the same kebab-case normalization from `personaFormScript.ts` (`.trim().toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/^-|-$/g, '')`) inline on the `.map()` step, keeping wizard-created and manually-created personas consistent regardless of what the model returns.

PR #60, opened 2026-06-07
Improve how multiagent agent gathers pr data

Sejdii:


.github/workflows/npm.yml
```diff
@@ -26,15 +26,7 @@ jobs:
...
-        os: [ubuntu-latest]
-    runs-on: ${{ matrix.os }}
-    steps:
-      - name: Publish
-        run: npm run deploy
-        env:
-          VSCE_PAT: ${{ secrets.VSCE_PAT }}
\ No newline at end of file
+    - run: npm run deploy
+      if: success() && startsWith(github.ref, 'refs/tags/')
```
Sejdii: **[AI Generated by CI/CD Expert persona]**
**Promotion gate removed — build and publish now share a job**

The original two-job layout (`build` → `publish` with `needs: [build]`) provided an explicit promotion boundary: all build-matrix legs had to succeed before publish ran. Collapsing publish into the build job removes that gate — a tag-triggered deploy can now begin on one OS runner even if other matrix runners are still running or have failed.

Restore a separate job with `needs: [build]` to re-establish the gate.

readme.md
```diff
@@ -56,6 +56,12 @@ No API keys are needed in Reviseo itself — it generates prompts, not AI respon
 
 ---
 
+## Working Directory
+
+In multi-agent mode, the orchestrator uses `.ai/reviseo/` as a working directory to store temporary review data files. Add `.ai/` to your repository's `.gitignore` to avoid committing these files.
```
Sejdii: **[AI Generated by CI/CD Expert persona]**
**`.gitignore` update not included in this PR**

The readme now advises users to add `.ai/` to `.gitignore`, but the PR ships no `.gitignore` change. Temporary files written to `.ai/reviseo/{reviewNumber}/review_data.md` contain the full PR diff; if a developer forgets, those files — potentially containing credentials embedded in diffs — get committed. Ship the `.gitignore` update in the same PR as the documentation that mandates it.

Sejdii: **[AI Generated by Skeptical Architect persona]**
**No cleanup lifecycle for working-directory temp files**

The README advises adding `.ai/` to `.gitignore` but says nothing about when or how these files are removed. Every review run writes persistent files that accumulate indefinitely. For a tool that could be run against many PRs this is unbounded disk growth with no GC strategy. Temporary shared state almost always needs a defined owner for cleanup — that owner is absent here. Consider either: (1) having the orchestrator delete the file on completion, or (2) documenting the expected lifecycle explicitly.

src/core/review-generation/review/MultiAgentPhaseComponent.ts
```diff
@@ -6,17 +6,29 @@ export class MultiAgentPhaseComponent extends PhaseComponent {
 `## Review Process
 You are running in multi-agent orchestration mode. Your role is to coordinate subagents — you do not perform code review yourself.
 
-1. Fetch the pull request diff using available tools.
-   Store the diff — it will be passed to all subagents.
-
-2. For each persona step below, launch a subagent with:
-   - The full PR diff.
+1. Extract the PR number from the pull request URL in the system prompt.
+   Derive the review data file path: \`.ai/reviseo/{reviewNumber}/review_data.md\`
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**No failure mode specified for step 1**

Every subsequent step ends with "If any step fails, report the error and stop." Step 1 — extracting the PR number from the URL — has no equivalent guard. If the system prompt contains no URL or has an unexpected format, the model has no instruction and may silently invent a number or continue with a malformed path.

Add: *"If the PR URL is not present or the number cannot be parsed, report the error and stop."*

Sejdii: **[AI Generated by Skeptical Architect persona]**
**LLM-based URL parsing belongs in code, not the prompt**

Instructing the LLM to "extract the PR number from the pull request URL in the system prompt" is prompt-engineering a string-split operation. PR URL parsing is deterministic, one-line TypeScript. Delegating it to an LLM introduces an unnecessary failure mode (hallucinated number, wrong format) and moves logic that belongs in the extension layer into natural-language instructions.

Extract the PR number in `extension.ts` before constructing the prompt, and inject it as a concrete value.

src/core/review-generation/review/MultiAgentPhaseComponent.ts
```diff
@@ -6,17 +6,29 @@ export class MultiAgentPhaseComponent extends PhaseComponent {
...
+
```
Sejdii, line 16: **[AI Generated by Prompt Engineer persona]**
**`{reviewNumber}` looks like an unresolved template token**

The literal string `{reviewNumber}` will appear verbatim in the generated prompt. An LLM reading this may treat it as a Jinja/Handlebars placeholder that was never substituted, rather than an instruction to substitute the value it just extracted. This creates a real risk the model uses the literal path `.ai/reviseo/{reviewNumber}/review_data.md` or becomes confused.

Either have step 1 explicitly say *"replace `{reviewNumber}` in the path below with the PR number you extracted"*, or change the notation to something that reads as a slot to fill, e.g. `<pr-number>`.

Sejdii, line 16: **[AI Generated by Clean Code Expert persona]**
**File path repeated three times — DRY violation in template string**

The path `.ai/reviseo/{reviewNumber}/review_data.md` is written verbatim on three separate lines (here, in step 2's writing instruction, and in step 3's subagent brief). If the directory layout changes, all three occurrences must be updated in sync.

Consider defining the path once at the top of step 1 — e.g. *"call this the review data file"* — and referencing it by that label in subsequent steps.

src/core/review-generation/review/MultiAgentPhaseComponent.ts
```diff
@@ -6,17 +6,29 @@ export class MultiAgentPhaseComponent extends PhaseComponent {
...
+2. Launch a data preparation subagent tasked only with:
+   - Fetching PR details via \`pull_request_read\` with method "get" (to obtain title and description)
+   - Fetching the diff via \`pull_request_read\` with method "get_diff"
+   - Writing \`.ai/reviseo/{reviewNumber}/review_data.md\` with the following sections:
+     # Title
+     <PR title>
```
Sejdii: **[AI Generated by Prompt Engineer + Skeptical Architect personas]**
**Bespoke Markdown schema should be structured data**

Two compounded issues with the file format:

1. **Fragile for the writing model**: the expected layout (`# Title`, `# Description`, `# Diff`) is specified as indented headings inside prose. The writing model must infer these are literal Markdown headings to output verbatim, not structural labels. A fenced code block showing the exact expected content, or an explicit instruction *"write the file with the following verbatim Markdown structure:"*, removes the ambiguity.

2. **Wrong format category**: Markdown is a presentation format, not a data interchange format. When the reader subagent parses `review_data.md`, it must infer structure from prose headings — adding fragility that a `review_data.json` with a `title`, `description`, and `diff` field would eliminate entirely. JSON/YAML gives parsers, schema validation, and unambiguous structure at zero extra cost.

src/core/review-generation/review/MultiAgentPhaseComponent.ts
```diff
@@ -6,17 +6,29 @@ export class MultiAgentPhaseComponent extends PhaseComponent {
...
+     # Description
+     <PR description>
+     # Diff
+     <full diff>
+   If any step fails, report the error and stop.
+
```
Sejdii, line 27: **[AI Generated by Prompt Engineer persona]**
**Failure instruction is scoped to the subagent, not the orchestrator**

The "If any step fails, report the error and stop" instruction lives inside the data-prep subagent's task description. The orchestrator, however, is the one who decides whether to continue to step 3. Without an explicit orchestrator-level instruction, the model may launch reviewer subagents regardless of what the data-prep subagent returned.

Add an instruction at the orchestrator level: *"If the data preparation subagent reports an error or does not confirm a successful file write, stop and report the error."*

src/core/review-generation/review/MultiAgentPhaseComponent.ts
```diff
@@ -6,17 +6,29 @@ export class MultiAgentPhaseComponent extends PhaseComponent {
...
+3. For each persona step below, launch a subagent with:
+   - The file path \`.ai/reviseo/{reviewNumber}/review_data.md\` — the subagent must read this file to obtain the PR context and diff.
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**Sequential dependency between step 2 and step 3 is only implied**

Step numbering implies ordering, but LLMs in agentic frameworks can parallelize or reorder steps if it seems efficient. The dependency is load-bearing: if reviewer subagents run before the file is written, they fail to read it.

Make the ordering explicit: *"Wait for the data preparation subagent to complete successfully before launching any reviewer subagents."*

src/core/review-generation/review/MultiAgentPhaseComponent.ts
```diff
@@ -6,17 +6,29 @@ export class MultiAgentPhaseComponent extends PhaseComponent {
...
    - That persona's instructions and checklist.
    - The JSON output format specified in the Final Step.
 
    If any subagent call fails, report the error and stop.
 
-3. Each subagent must return ONLY a JSON array. No prose, no markdown fences.
+4. Each subagent must return ONLY a JSON array. No prose, no markdown fences.
    Every element must conform to:
```
Sejdii, line 39: **[AI Generated by Prompt Engineer persona]**
**"Each subagent" is ambiguous — data-prep subagent is also a subagent**

Step 4 says "Each subagent must return ONLY a JSON array", but the data preparation subagent launched in step 2 is also a subagent and is not expected to return a JSON array — it writes a file. The current wording could cause the model to misapply the JSON constraint to the data-prep subagent, or expect a JSON array back from it before proceeding.

Qualify explicitly: *"Each reviewer subagent (launched in step 3) must return ONLY a JSON array."*

Sejdii (commented):
Multi-agent code review (4 personas: CI/CD Expert, Clean Code Expert, Prompt Engineer, Skeptical Architect).

**Summary of findings:**

**CI/CD (3 high-priority):** The publish step was collapsed into the build job, removing the `needs: [build]` promotion gate and introducing a risk of duplicate publishes across matrix OS runners. A missing post-publish smoke test and the absence of a `.gitignore` update (despite the readme requiring it) round out the pipeline concerns.

**Prompt Engineering (6 findings):** The `{reviewNumber}` placeholder reads as an unresolved template token to an LLM. Several structural issues: no failure guard on step 1, sequential dependency between steps 2 and 3 is only implied, the "each subagent" JSON constraint is ambiguous (it also applies to the data-prep subagent by current wording), and no tool name is given for reviewer subagents to read the shared file.

**Architecture (2 findings):** The filesystem-as-IPC pattern with a dedicated data-prep subagent adds unnecessary complexity — the orchestrator could fetch the diff inline. The bespoke Markdown schema (`# Title / # Description / # Diff`) is fragile; JSON would be unambiguous and machine-parseable.

**Clean Code (2 findings):** The file path is repeated verbatim three times (DRY violation). `success()` in the deploy step condition is redundant and adds reading noise.

PR #62, opened 2026-06-08
Improve code review prompt

Sejdii:


src/core/review-generation/PromptBuilder.ts
```diff
@@ -34,10 +46,22 @@ export class PromptBuilder {
...
 
-        return new Prompt([new SystemPromptComponent(this._url), phase]);
+        const reviewPhaseFactory = (n: number): PromptComponent => this._mode === Modes.MULTI_AGENT
+            ? new MultiAgentProvideMultipersonaReviewPhase(n, stepsComponent, prNumber)
+            : new SingleAgentProvideMultipersonaReviewPhase(n, stepsComponent, prNumber);
+
+        const phaseFactories: Array<(n: number) => PromptComponent> = [
+            n => new FetchReviewDataPhase(n, prNumber),
+            reviewPhaseFactory,
+            n => new ValidateCommentsPhase(n),
```
Sejdii: **[AI Generated by Clean Code Expert persona]**
**`phaseFactories` array mixes extracted and inline factories inconsistently**

`reviewPhaseFactory` is extracted to a named variable while the other four phases use inline arrow functions. The asymmetry makes readers ask "why is this one special?" — the answer (it needs a conditional) is valid, but not communicated by the structure.

Either extract all phase constructors to named variables, or document why this one is special with an inline comment. The current hybrid implies a false distinction.

src/core/review-generation/review/ProvideMultipersonaReviewPhase.ts
```diff
@@ -0,0 +1,32 @@
+import type { PromptComponent } from './PromptComponent';
+import { CommentFormatComponent } from './CommentFormatComponent';
+import type { StepsComponent } from './StepsComponent';
+
+export const REVIEW_ASSUMPTIONS =
```
Sejdii: **[AI Generated by Clean Code Expert + Software Architect personas]**
**`REVIEW_ASSUMPTIONS` exported constant and abstract base both leaked on module surface**

Two related issues here:

1. `REVIEW_ASSUMPTIONS` is a plain string constant exported from a file named after an abstract class. Consumers importing it must depend on an unrelated file. If it's shared, it belongs in its own file or in `CommentFormatComponent`; if it's only used internally, it should be unexported.

2. `ProvideMultipersonaReviewPhase` (the abstract base) is re-exported in `review/index.ts`. External consumers of the `review/` module should depend on concrete types or the `PromptComponent` interface — not the inheritance hierarchy. Exposing the abstract base leaks an implementation detail and creates a surface that callers could import and extend in ways not intended. It should remain internal to the `review/` folder.

src/core/review-generation/review/OrchestratorSystemPromptComponent.ts
```diff
@@ -0,0 +1,14 @@
+import type { PromptComponent } from './PromptComponent';
+
+const ROLE =
+`You are an orchestrator coordinating a structured, multi-phase code review process.
+Your role is to execute each phase in order — each phase must be handled by a dedicated subagent.
+Launch one subagent per phase, wait for it to complete successfully, then proceed to the next.
+You do not perform code review yourself — you delegate that work to the appropriate phase.
+---`;
+
```
Sejdii, line 8: **[AI Generated by Prompt Engineer persona]**
**`---` separator's purpose is implicit — model receives no instruction on what it means**

The role text ends with `---`, a visual separator carried over from the old `SystemPromptComponent`. The model is never told what the separator signals or that phases follow it. Since the `## Phase N:` headings already provide clear structural markers, the bare `---` is redundant noise. Either remove it, or replace it with something explicit like `--- Begin phases ---` if a separator is genuinely needed here.

src/core/review-generation/review/SingleAgentProvideMultipersonaReviewPhase.ts
```diff
@@ -0,0 +1,14 @@
+import { ProvideMultipersonaReviewPhase } from './ProvideMultipersonaReviewPhase';
+
+export class SingleAgentProvideMultipersonaReviewPhase extends ProvideMultipersonaReviewPhase {
+    protected buildInstructions(): string {
+        const reviewDataPath = `.ai/reviseo/${this.prNumber}/review_data.json`;
+        return `## Phase ${this.phaseNumber}: Provide Multipersona Review
+
+Read \`${reviewDataPath}\` to obtain the PR context and diff.
+For each Step below, adopt that persona and perform your review according to its checklist.
+Generate review comments but do NOT submit them — submission happens in a later phase.
```
Sejdii, line 10: **[AI Generated by Prompt Engineer persona]**
**Single-agent mode gives no internal format for comments between phases**

The instruction says "Generate review comments but do NOT submit them — submission happens in a later phase." But there's no guidance on the format the model should use to hold those comments in memory between phases. The multi-agent path specifies a JSON schema for inter-agent communication; the single-agent path leaves the format undefined.

Without a defined format, the `ValidateCommentsPhase` and `SubmitReviewPhase` have no stable structure to work with. Consider adding something like: "Retain comments as an in-memory list; each item: persona name, file path, line number, title, body."

src/core/review-generation/review/ValidateCommentsPhase.ts
```diff
@@ -0,0 +1,30 @@
...
+2. Merge such comments into a **single, stronger comment**:
+   - Preserve the most insightful reasoning from each.
+   - Mention all relevant personas in the header, e.g.
+     \`**[AI Generated by Software Architect + Security Reviewer personas]**\`
+     \`**Memory Leak in Connection Pool**\`
+   - Use the most specific file/line reference.
+   - A comment is redundant only if its entire body is already covered by the merged comment.
+   - When in doubt, keep both comments rather than dropping one.
+3. Remove any remaining redundant or low-value comments.
+4. Verify that every remaining comment conforms to the comment format:`;
```
Sejdii, line 23: **[AI Generated by Prompt Engineer persona]**
**Phase ends without a closing action — model doesn't know what "done" looks like**

The phase text ends after point 4 ("Verify that every remaining comment conforms to the comment format:") followed by the format block. There's no sentence telling the model what output it should produce or what signals phase completion.

Without a closing instruction, the model may not know when this phase ends or what to hand off to Phase 4. Adding something like: "Once validated, retain the final comment list for submission in the next phase." closes the loop and makes the handoff explicit.

src/core/review-generation/review/FetchReviewDataPhase.ts
```diff
@@ -0,0 +1,27 @@
...
+2. Fetch the diff via \`pull_request_read\` with method "get_diff"
+3. Write \`${reviewDataPath}\` as a JSON file with this exact structure:
+   {
+     "title": "<PR title>",
+     "description": "<PR description>",
+     "diff": "<full diff>"
+   }
+
+If any step fails, report the error and stop.`;
+    }
```
Sejdii, line 32: **[AI Generated by Prompt Engineer persona]**
**Missing explicit write-confirmation instruction before signaling phase complete**

The old multi-agent implementation included: "Wait for the data preparation subagent to complete successfully before proceeding. If it reports an error or does not confirm a successful file write, stop and report the error."

The new `FetchReviewDataPhase` only says "If any step fails, report the error and stop." — it doesn't instruct the model to confirm the file was actually written before reporting success. In an orchestrator that delegates phases to subagents, Phase 2 could start before the file exists if the subagent fails silently or partially.

Consider adding: "Confirm the file has been written successfully before reporting this phase complete."

src/core/review-generation/review/FetchReviewDataPhase.ts
```diff
@@ -0,0 +1,27 @@
...
+    constructor(
+        private readonly phaseNumber: number,
+        private readonly prNumber: string,
+    ) {}
+
+    getText(): string {
+        const reviewDataPath = `.ai/reviseo/${this.prNumber}/review_data.json`;
+        return `## Phase ${this.phaseNumber}: Fetch Review Data
+
+The review data file for this review is: \`${reviewDataPath}\`
```
Sejdii, line 18: this phase should receive whole url of pull request

PR #65, opened 2026-06-11
Add option to leave review in pending state

Sejdii:


src/core/review-generation/PromptBuilder.ts
```diff
@@ -20,22 +20,27 @@ export class PromptBuilder {
         private readonly _personas: readonly Persona[] = [],
         private readonly _mode = Modes.SINGLE_AGENT,
         private readonly _context: Record<string, Record<string, string>> = {},
+        private readonly _pendingReview = false,
     ) {}
 
     url(prUrl: string): PromptBuilder {
-        return new PromptBuilder(prUrl, this._personas, this._mode, this._context);
+        return new PromptBuilder(prUrl, this._personas, this._mode, this._context, this._pendingReview);
```
Sejdii: **[AI Generated by Clean Code Expert persona]**
**Telescoping constructor forces all setter methods to be updated for every new field**

Every setter method (`url`, `personas`, `context`, `pendingReview`) must now pass all five constructor arguments when producing a new `PromptBuilder` instance. Adding `_pendingReview` in this PR already required touching every existing method. The next new field will require the same across all six methods (including `mode`), and a mistake — forgetting to thread one argument — produces a silent reset-to-default bug with no compile-time warning.

The underlying problem is that state is spread across discrete constructor parameters rather than a single state record. A minimal fix:

```typescript
interface BuilderState {
    url: string;
    personas: readonly Persona[];
    mode: Modes;
    context: Record<string, Record<string, string>>;
    pendingReview: boolean;
}

private constructor(private readonly _state: BuilderState) {}

url(prUrl: string): PromptBuilder {
    return new PromptBuilder({ ...this._state, url: prUrl });
}
```

With this shape, adding a new field touches: the state type, one new setter method, and the `build()` method — and nothing else. The call sites remain identical.

.claude/skills/review-apply/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+# review-apply
+
+Fetch all review comments from a pull request, decide how to handle each one, apply code changes where appropriate, and reply to each comment on GitHub.
+
+## Input
+
+The user provides a PR reference in one of these forms:
+- A PR number: `/review-apply 42`
+- A full GitHub URL: `/review-apply https://github.com/owner/repo/pull/42`
+- Just a number with context implied by the current repo: `/review-apply 42`
```
Sejdii: **[AI Generated by Clean Code Expert persona]**
**Duplicate input bullet obscures what input forms are actually supported**

The Input section lists three forms, but bullets 1 and 3 are identical:

```
- A PR number: `/review-apply 42`
- A full GitHub URL: `/review-apply https://github.com/owner/repo/pull/42`
- Just a number with context implied by the current repo: `/review-apply 42`
```

The apparent intent of bullet 3 is to clarify that the current repo's git remote is used when no owner/repo is explicit. But since the example is the same as bullet 1, a reader (or the model following this skill) cannot tell whether these are two genuinely distinct cases or a copy-paste error.

Suggestion: collapse to two bullets and move the "context implied by current repo" explanation inline:

```
- A PR number (owner/repo inferred from the current git remote): `/review-apply 42`
- A full GitHub URL: `/review-apply https://github.com/owner/repo/pull/42`
```

src/core/review-generation/review/SubmitReviewPhase.ts
```diff
@@ -4,16 +4,25 @@ export class SubmitReviewPhase implements PromptComponent {
...
     ) {}
 
     getText(): string {
-        return `## Phase ${this.phaseNumber}: Submit Review
+        const title = this.pendingReview ? 'Create Pending Review' : 'Submit Review';
+        const header = this.pendingReview
+            ? `Create a pending review on pull request #${this.prNumber} and add all validated comments. Do NOT submit it:`
+            : `Create a pending review on pull request #${this.prNumber} and add all validated comments:`;
+        const stopStep = this.pendingReview
+            ? '\n3. Stop here — do NOT submit or publish the review. Leave it in pending/draft state for manual inspection before publishing.'
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**Critical "stop" constraint placed as the last numbered step, where it is least likely to constrain model behavior**

In the pending-review path, the prohibition on submitting is emitted as step 3 in an ordered list:

```
1. Create a new pending review...
2. Add each validated comment...
3. Stop here — do NOT submit or publish the review.
```

Models execute ordered instructions sequentially. The constraint arrives only after steps 1 and 2 are complete, meaning the model has finished acting before it reads the stop condition. Worse, if the model anticipates what follows a "add comments to review" step from training data (i.e., "submit the review"), it may act before it reaches step 3 at all.

The phase header already contains "Do NOT submit it:" — that's the right location for this constraint. The stopStep text should be folded into the header and removed as a numbered step:

```
Create a pending review on pull request #N, add all validated comments, and stop — do NOT submit or publish it. Leave it in pending/draft state for manual inspection.
```

Then steps 1 and 2 remain as the only numbered actions, with no ambiguous "step 3" that implies continuation after step 2.

.claude/skills/review-apply/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+- PR title, base branch, head branch
+- All review comments (inline comments on specific lines/files)
+- All review threads
+
+Also use `mcp__github__list_pull_requests` if needed to confirm the PR state.
+
+Collect every comment that:
+- Is unresolved (not marked as resolved/outdated)
+- Is an inline code comment (has a `path` and `position`/`line`)
+- Or is a general review comment requesting changes
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**Ambiguous "Or" in comment filter may cause resolved general comments to be collected**

The collection filter reads:

```
Collect every comment that:
- Is unresolved (not marked as resolved/outdated)
- Is an inline code comment (has a `path` and `position`/`line`)
- Or is a general review comment requesting changes
```

The "Or" on the third bullet introduces a structural ambiguity: does "unresolved" (bullet 1) apply to the third bullet? A model can reasonably parse this as either:
- `(unresolved AND inline) OR (any general comment)` — collects resolved general comments, the likely unintended reading
- `unresolved AND (inline OR general)` — the intended interpretation

Make the scope of "unresolved" unambiguous by restructuring:

```
Collect every comment that is unresolved (not marked as resolved/outdated) AND is one of:
- An inline code comment (has a `path` and `position`/`line`)
- A general review comment requesting changes
```

src/extension/webview/ReviseoPanel.ts
```diff
@@ -9,11 +9,14 @@ import type { WebviewTab } from './WebviewTab';
 import { PersonasTab } from './tabs/personasTab';
 import { ReviewTab } from './tabs/reviewTab';
 
+const REVIEW_SETTINGS_KEY = 'reviseo.reviewSettings';
```
Sejdii, line 12: **[AI Generated by Skeptical Architect persona]**
**globalState used for user preferences when VSCode's configuration contribution system already solves this**

Settings are persisted via `globalState.get/update` with a manually managed `REVIEW_SETTINGS_KEY`. VSCode provides `contributes.configuration` in `package.json` precisely for this use case, which gives:

- Declarative schema with types, defaults, and descriptions in one place
- Built-in user-scoped persistence via `vscode.workspace.getConfiguration()`
- Automatic exposure in the VSCode Settings UI (searchable, resettable, documented)
- No custom messaging protocol needed — the extension reads settings directly without a webview round-trip

The current implementation adds: a constant key, a custom `getReviewSettings` message type, a `reviewSettingsLoaded` response type, a `saveReviewSettings` message type, a dispatcher in `shared.ts`, an init request in `shared.ts`, and two handler cases in `ReviseoPanel`. All of this is boilerplate that `getConfiguration()` + `contributes.configuration` eliminates.

The only tradeoff: `globalState` settings survive workspace changes, while `workspace.getConfiguration()` can be scoped. For checkboxes like "leave review pending", user scope is appropriate and is directly supported.

src/core/review-generation/PromptBuilder.ts
Sejdii: Fixed — replaced the discrete constructor parameters with a single `BuilderState` interface; every setter now uses `{ ...this._state, field: value }` so adding a new field only requires touching the interface, the new setter, and `build()`. The constructor is now private with a `PromptBuilder.create()` static factory replacing `new PromptBuilder()`.

.claude/skills/review-apply/SKILL.md
Sejdii: Fixed — collapsed to two bullets with the "owner/repo inferred from current git remote" clarification inline on the first bullet.

src/core/review-generation/review/SubmitReviewPhase.ts
Sejdii: Fixed — folded the stop constraint into the header ("…and stop — do NOT submit or publish it. Leave it in pending/draft state for manual inspection.") and removed `stopStep` as a numbered step, leaving only steps 1 and 2. The fixture file was updated to match.

.claude/skills/review-apply/SKILL.md
Sejdii: Fixed — restructured to "Collect every comment that is unresolved … AND is one of: …" so the scope of "unresolved" unambiguously covers both inline and general comments.

src/extension/webview/ReviseoPanel.ts
Sejdii: Will not be applied — the webview runs in a sandboxed iframe and cannot call VSCode APIs directly, so a messaging protocol between the webview and the extension host is unavoidable regardless of whether settings are stored in `globalState` or `contributes.configuration`. Migrating to `getConfiguration()` would still require all the same message types to push values into the webview and receive saves back, while also adding a `contributes.configuration` block to `package.json` and a `onDidChangeConfiguration` listener. The net complexity is higher, not lower, for this use case. This is worth discussing as a separate PR if the Settings UI exposure is a goal.

PR #66, opened 2026-06-13
Refactor prompt builder to abstract factory

Sejdii:


src/core/review-generation/ReviewConfiguration.ts
```diff
@@ -0,0 +1,60 @@
+import type { Persona } from '../persona-management';
```
Sejdii, line 1: **[AI Generated by Clean Code Expert + Software Architect personas]**
**Public constructor and builder both exported creates ambiguous construction API**

`ReviewConfiguration` exposes a public constructor (all five fields positional) while also providing `static builder()`. This gives consumers two unguided construction paths. If the builder is the intended API, the constructor should be `private` to enforce it — otherwise contributors will mix both styles over time.

A related concern at the module boundary: `ReviewConfigurationBuilder` is exported from `index.ts` alongside `ReviewConfiguration`. The builder is a construction detail, not a public contract. If it's renamed or refactored internally, all external import sites break. The intended external API should be `ReviewConfiguration` (accessed via `ReviewConfiguration.builder()`) — the builder class itself does not need to be a named export.

Similarly, `GithubReviewPromptFactory` is exported from `index.ts` even though the `ReviewGenerationFacade` already defaults to it internally. Exporting the concrete class invites consumers to depend on it directly rather than on the `ReviewPromptFactory` interface, which undermines the DIP benefit this refactor was designed to establish.

Sejdii: Fixed — all three concerns addressed:

1. **Private constructor**: `ReviewConfiguration` constructor is now `private`. To keep the builder working across class boundaries, `ReviewConfiguration.builder()` passes a factory closure into `ReviewConfigurationBuilder`, so the builder calls `new ReviewConfiguration(...)` through a captured closure (the only valid call site) rather than directly — enforcing the private constraint at the TypeScript level.

2. **`ReviewConfigurationBuilder` removed from `index.ts`**: it's a construction detail accessible only via `ReviewConfiguration.builder()`, and the class itself is no longer exported from the module at all (not even from the file).

3. **`GithubReviewPromptFactory` removed from `index.ts`**: external consumers should depend on the `ReviewPromptFactory` interface, not the concrete class. The facade already defaults to it internally, so there's no reason to expose it at the module boundary.

PR #68, opened 2026-06-14
Add local review

Sejdii:


src/core/review-generation/ReviewConfiguration.ts
```diff
@@ -61,13 +67,25 @@ class ReviewConfigurationBuilder {
...
 
+    withBaseBranch(baseBranch: string): this {
+        this._baseBranch = baseBranch;
+        return this;
+    }
+
+    withTimestamp(timestamp: string): this {
+        this._timestamp = timestamp;
+        return this;
+    }
```
Sejdii: **[AI Generated by Software Architect + Skeptical Architect personas]**
**ReviewConfiguration conflates two distinct review contexts and silently accepts broken field combinations**

`ReviewConfiguration` now carries `url` and `pendingReview` (GitHub-specific) alongside `baseBranch` and `timestamp` (local-specific). Depending on the active review mode, half of these fields are meaningless — yet both sets are always present and default to empty strings. The class now has an implicit discriminated union baked into it without expressing the discrimination.

This is a classic symptom of extending a working DTO under pressure rather than introducing a proper type boundary. Any reader of `ReviewConfiguration` must mentally track which fields are valid in which context, and the compiler offers no protection against contradictory combinations (e.g. a URL and a timestamp set simultaneously). When a third review target (GitLab, Bitbucket) arrives, this pattern repeats and the class becomes a kitchen-sink.

The builder compounds this: `_baseBranch` and `_timestamp` are initialised to `''`. If `buildLocalPrompt` is called without setting these — which TypeScript cannot currently prevent — `LocalReviewDiffFilePath` will generate paths like `.ai/reviseo//local.diff` with no error at the call site.

**Proven alternatives:**
1. **Discriminated union** — `type ReviewConfiguration = { kind: 'github'; url: string; pendingReview: boolean; personas: ...; ... } | { kind: 'local'; baseBranch: string; timestamp: string; personas: ...; ... }`. This makes illegal states unrepresentable, lets each factory accept only the shape it actually needs, and allows TypeScript exhaustive checks on `kind` to propagate safety through the whole pipeline at compile time.
2. **Separate configuration interfaces** — `GithubReviewConfiguration` and `LocalReviewConfiguration` both implement a minimal shared `ReviewConfigurationBase` containing only `personas`, `personaReviewExecutionMode`, and `context`. Each factory receives only the type it understands.
3. **Runtime guard as a stopgap** — Add a validation check in `build()` that throws when required fields for the active configuration mode are absent, or use a typed step-builder chain to make `withBaseBranch` and `withTimestamp` mandatory at the type level.

Neither option 1 nor 2 allows a caller to populate nonsensical cross-type fields. Adding a third review target requires adding a third type — not patching an existing one.

src/core/review-generation/ReviewGenerationFacade.ts
Sejdii, line 1: **[AI Generated by Software Architect + Skeptical Architect personas]**
**Facade API grows linearly with review targets instead of using polymorphic dispatch — OCP violation**

The facade now holds two concrete factories (`_githubFactory`, `_localFactory`) and exposes two separate public methods (`buildPrompt`, `buildLocalPrompt`). The selection of which factory to invoke is delegated entirely to the caller — the facade itself has no dispatch logic. This is the Open/Closed Principle inverted: adding a third review source (e.g. a local diff of a stash, a GitLab MR) requires modifying the facade by adding a third factory field and a third method.

The original design, where the facade accepted a single `ReviewPromptFactory`, was actually more open for extension — the factory was injectable. This PR inlines both concrete factories as defaults, removing that flexibility. The routing logic (which factory, which method) now leaks into the UI layer (`ReviseoPanel`), which is the wrong place for it.

**Proven alternatives:**
1. **Strategy pattern with a discriminated input** — A single `build(config: ReviewConfiguration): string` method on the facade selects the factory internally based on `config.kind`. New targets only require registering a new factory; no facade interface change:
```ts
buildPrompt(config: ReviewConfiguration): string {
    const factory = config.kind === 'github' ? this._githubFactory : this._localFactory;
    return factory.create(config).getText();
}
```
2. **Factory registry** — `Map<ReviewKind, ReviewPromptFactory>` keyed by the discriminant. The facade looks up the factory at runtime: `this._factories.get(config.kind).create(config).getText()`. Each new target is one `set()` call on construction.

Both approaches keep the public API surface stable as the set of review targets grows. The UI layer only needs to know 'what kind of review' — not 'which facade method'. The facade stays closed to modification and open to extension.

src/core/review-generation/LocalReviewDiffFilePath.ts
Sejdii, line 1: **[AI Generated by Skeptical Architect persona]**
**LocalReviewDiffFilePath is a class wrapping two string concatenations**

`LocalReviewDiffFilePath` is a class with a constructor accepting a `timestamp` and two methods that return template-literal strings. It carries no invariants (any string is accepted), no validation, and no behaviour beyond computing two derived strings from a single input.

This is a textbook example of an anemic value object — a class that exists only to group two property accesses that could just as easily be plain functions or a simple record literal. Its presence adds a named type to the dependency graph of `LocalReviewPromptFactory`, `LocalFetchDiffPhase`, and `LocalCleanupPhase`, all of which must now import it to obtain what is ultimately a pair of strings.

**Why this matters:** When a pattern like this establishes itself in a codebase, it tends to propagate. The next path type (`GithubReviewDataFilePath`? `LocalReviewOutputFilePath`?) follows the same template, and the type graph expands with zero semantic gain.

**Proven alternatives:**
1. **Plain helper function** — `function localReviewPaths(timestamp: string) { return { fullPath: ..., directory: ... }; }` — a named, testable function that returns a typed record. No class, no instantiation ceremony, fully tree-shakeable.
2. **Inline the derivation at the single call site** — `LocalReviewPromptFactory` creates both strings directly from `config.timestamp` and passes them as plain strings. If the path format ever changes, it changes in one place.

A function or inline derivation is the right tool for a pure string computation. A class adds object identity, instantiation cost, and import coupling for no benefit when there is no mutable state, no interface to implement, and no polymorphism needed.

src/core/review-generation/GithubReviewPromptFactory.ts
```diff
@@ -22,13 +22,14 @@ export class GithubReviewPromptFactory implements ReviewPromptFactory {
         }
 
         const prNumber = config.url.match(/\/pull\/(\d+)/)?.[1] ?? '0';
```
Sejdii, line 25: **[AI Generated by Software Architect persona]**
**`.ai/reviseo/` path convention duplicated between two independent sites**

`GithubReviewPromptFactory` constructs `.ai/reviseo/${prNumber}/review_data.json` inline, while `LocalReviewDiffFilePath` independently encodes the same `.ai/reviseo/` root. These two sites will diverge silently if the output directory is ever reconfigured.

Extract a shared `ReviseoWorkspaceLayout` (or similar) in `core/` that owns the canonical path shapes for both review sources. Both factories consume it, and the constant lives in exactly one place. An architecture test should assert that no string literal matching `\.ai/reviseo/` appears outside that one module.

src/extension/webview/ReviseoPanel.ts
```diff
@@ -113,6 +132,10 @@ export class ReviseoPanel {
         return getWebviewHtml(nonce, cspSource, tabs);
```
Sejdii, line 144: **[AI Generated by Software Architect persona]**
**Timestamp format chosen in extension layer drives file path structure owned by core**

`_generateTimestamp()` in the VSCode extension layer decides the string format (`2026-06-14T10-30-00`) that becomes a directory name inside `LocalReviewDiffFilePath`. The format is a filesystem concern (colons are illegal on Windows paths) that is tightly coupled to how `core/` constructs paths — yet the decision is made outside core.

The timestamp format contract should be defined in `core/` (e.g., as a factory method or documented type alias on `LocalReviewDiffFilePath`), and the extension layer should call that factory rather than rolling its own regex. This keeps the layer boundary clean: core owns the path structure, the extension layer merely supplies a raw `Date`.

src/core/review-generation/ReviewConfiguration.ts
Sejdii: Will not be applied — the project has an established pattern of a single `ReviewConfiguration` (decided in a prior refactor to avoid subclassing per review mode); a discriminated union would contradict that decision and require broader changes across the facade and all callers.

src/core/review-generation/ReviewGenerationFacade.ts
Sejdii: Will not be applied — unified dispatch via `config.kind` requires a discriminated union on `ReviewConfiguration`, which conflicts with the project's established single-config pattern; with only two review modes the two explicit methods are clear and the OCP trade-off is not yet worth the architectural coupling.

src/core/review-generation/LocalReviewDiffFilePath.ts
Sejdii: Fixed — converted `LocalReviewDiffFilePath` from a class to a plain `localReviewPaths(timestamp)` function returning `{ fullPath, directory }`; updated `LocalFetchDiffPhase` and `LocalCleanupPhase` to accept plain strings, and `LocalReviewPromptFactory` to call the function directly.

src/core/review-generation/GithubReviewPromptFactory.ts
Sejdii: Partially fixed — extracted a shared `REVISEO_BASE_DIR` constant from `LocalReviewDiffFilePath.ts` and used it in both `GithubReviewPromptFactory` and `localReviewPaths`; the suggested architecture test is out of scope for this PR.

src/extension/webview/ReviseoPanel.ts
Sejdii: Fixed — moved timestamp generation to `core/` as `generateLocalReviewTimestamp(date?: Date): string` exported from `review-generation/index.ts`; `ReviseoPanel` now imports and calls it directly, removing `_generateTimestamp()` from the extension layer.

src/core/review-generation/ReviewConfiguration.ts
Sejdii: Fixed (superseding earlier rejection) — `ReviewConfiguration` is now a discriminated union: `GithubReviewConfiguration` (`kind: 'github'`, `url`, `pendingReview`) | `LocalReviewConfiguration` (`kind: 'local'`, `baseBranch`, `timestamp`), with shared fields in a base type. The builder was removed; callers construct plain object literals. Both factories narrow on `config.kind` before accessing kind-specific fields.

src/core/review-generation/ReviewGenerationFacade.ts
Sejdii: Fixed (superseding earlier rejection) — `buildPrompt` and `buildLocalPrompt` replaced with a single `build(config: ReviewConfiguration): string` that dispatches to the right factory via `config.kind`. Adding a new review target now only requires registering a new factory; the facade interface stays closed to modification.

PR #69, opened 2026-06-20
Remember last used review mode per workspace

Sejdii:


src/extension/webview/tabs/reviewTab/reviewScript.ts
```diff
@@ -11,6 +11,12 @@ export function getReviewScript(): string {
       document.getElementById('chk-multi-agent').checked = e.detail.multiAgent;
       document.getElementById('base-branch').value = e.detail.baseBranch;
     });
+    document.addEventListener('active-review-tab-loaded', (e) => {
+      const btn = document.querySelector('.inner-tab-btn[data-inner-tab="' + e.detail.tab + '"]');
+      if (btn) { btn.click(); }
```
Sejdii: **[AI Generated by Clean Code Expert + Skeptical Architect + VS Code Extension Development Expert personas]**
**Tab restore via `btn.click()` fires unintended side effects on every panel open**

Restoring the active tab by calling `btn.click()` re-runs the full tab-switch handler, which always:
1. Posts `saveActiveReviewTab` back to `workspaceState` — a no-op write on every panel open.
2. Fires `getReviewSettings` or `getLocalReviewSettings` as a side-effect of *restoration*, not of a user action.

This entangles initialisation logic with user-interaction logic. The two intents — "restore saved state" and "user switched tab" — should be separate code paths.

**Recommended fix:** Extract the visual switch logic into a dedicated `activateTab(tab)` helper (update classes, toggle visibility, set `activeInnerTab`). The click handler calls `activateTab(tab)`, posts `saveActiveReviewTab`, and fetches settings. The restore listener calls only `activateTab(tab)` — no persistence, no settings fetch:

```js
function activateTab(tab) {
  activeInnerTab = tab;
  // update CSS classes, panel visibility, .github-only elements…
}

btn.addEventListener('click', () => {
  activateTab(btn.dataset.innerTab);
  vscode.postMessage({ type: 'saveActiveReviewTab', tab: activeInnerTab });
  loadSettingsForActiveTab();
});

document.addEventListener('active-review-tab-loaded', (e) => {
  activateTab(e.detail.tab); // restore only — no write, no fetch
});
```

Each path now has a single, clear responsibility, and the spurious write + round-trip on panel open are eliminated.

src/extension/webview/ReviseoPanel.ts
```diff
@@ -125,6 +126,15 @@ export class ReviseoPanel {
                 this._context.workspaceState.update(LOCAL_REVIEW_SETTINGS_KEY, { multiAgent: message.multiAgent, baseBranch: message.baseBranch });
                 break;
             }
+            case 'getActiveReviewTab': {
+                const tab = this._context.workspaceState.get<'github' | 'local'>(ACTIVE_REVIEW_TAB_KEY, 'github');
+                this._panel.webview.postMessage({ type: 'activeReviewTabLoaded', tab });
+                break;
+            }
+            case 'saveActiveReviewTab': {
+                this._context.workspaceState.update(ACTIVE_REVIEW_TAB_KEY, message.tab);
```
Sejdii: **[AI Generated by Skeptical Architect persona]**
**Growing get/save message-pair pattern — consider `retainContextWhenHidden` and/or bundled init state**

This PR adds a third independent get/save round-trip (`getActiveReviewTab` / `saveActiveReviewTab`) alongside the existing `getReviewSettings`/`saveReviewSettings` and `getLocalReviewSettings`/`saveLocalReviewSettings` pairs. Each new UI preference currently requires: a key constant, two switch-case branches in the panel, two message type declarations in `types.ts`, a `dispatchEvent` line in `shared.ts`, and call sites in the script. The team is incrementally building a DIY state-sync mini-framework.

There's also a UX consequence: `getActiveReviewTab` is posted after the DOM is ready, but the response arrives asynchronously. Between panel open and message receipt the UI renders with the hardcoded default (`activeInnerTab = 'github'`). Users who prefer the local tab will see a brief flash of the wrong tab on every panel open.

**Two platform-native alternatives:**

1. **`retainContextWhenHidden: true`** — keeps the webview process alive when the panel is hidden. The in-memory `activeInnerTab` variable already holds the correct value on show; zero custom code needed for ephemeral UI state like active tab. This is the idiomatic VS Code answer for lightweight UI preferences.

2. **Single bundled init message** — if cross-session persistence (panel fully closed and reopened) is required, a single `getInitialState` → `initialStateLoaded` message can carry *all* persisted values at once, injected synchronously into the HTML before first paint (via a `data-` attribute or inline `<script>` variable). This caps protocol growth at O(1) messages and eliminates the async flash entirely.

Either approach is a better fit than continuing to grow N parallel async round-trips.

Sejdii: **[AI Generated by VS Code Extension Development Expert persona]**
**`workspaceState.update()` Thenable not awaited — write failures are silently swallowed**

`this._context.workspaceState.update(ACTIVE_REVIEW_TAB_KEY, message.tab)` returns a `Thenable<void>` that is neither awaited nor `.then`-chained. If the underlying Memento write fails, the error is discarded silently and the webview receives no signal that the state was not persisted.

The same pattern exists in the pre-existing `saveLocalReviewSettings` handler, but this PR extends it. At minimum, attach a rejection handler so failures surface in the extension host output channel:

```ts
void this._context.workspaceState
  .update(ACTIVE_REVIEW_TAB_KEY, message.tab)
  .then(undefined, err => console.error('reviseo: failed to save active tab', err));
```

Or, since the surrounding `handleMessage` is already `async`, simply `await` the call and let any rejection propagate to the caller.

src/extension/webview/types.ts
Sejdii, line 1: **[AI Generated by Clean Code Expert persona]**
**`'github' | 'local'` union type repeated inline — missing domain alias**

The literal union `'github' | 'local'` appears in three separate places: twice in `types.ts` (once in `WebviewMessage`, once in `ExtensionMessage`) and a third time in `ReviseoPanel.ts` as a generic argument to `workspaceState.get<'github' | 'local'>`. There is no named alias capturing the domain concept.

The practical cost is fragility: if a third review mode is added, all three occurrences must be found and updated independently. The unnamed literal also provides no signal about *what* this value represents — a reader of `workspaceState.get<'github' | 'local'>` has to infer meaning from context.

Extracting a named export — e.g. `export type ReviewMode = 'github' | 'local'` in `types.ts` — makes the concept explicit, reduces the update surface to one location, and lets `ReviseoPanel.ts` reference `ReviewMode` instead of restating the union inline.

src/extension/webview/tabs/reviewTab/reviewScript.ts
Sejdii: Fixed — extracted `activateTab(tab)` to handle all visual state (classes, panel visibility, `.github-only` elements) and a separate `loadSettingsForActiveTab()` helper. The click handler calls all three; the `active-review-tab-loaded` restore listener calls only `activateTab` + `loadSettingsForActiveTab`, skipping the `saveActiveReviewTab` write.

src/extension/webview/ReviseoPanel.ts
Sejdii: Will not be applied — `retainContextWhenHidden: true` is already set in `ReviseoPanel.ts` (line 45), which is the idiomatic VS Code answer for this class of ephemeral UI state. The bundled init message approach is a valid long-term direction but is a larger refactor that goes beyond this PR's scope.

Sejdii: Fixed — the `saveActiveReviewTab` handler now uses `void` with a rejection handler to surface write failures in the extension host output channel instead of silently discarding them.

src/extension/webview/types.ts
Sejdii: Fixed — extracted `export type ReviewMode = 'github' | 'local'` in `types.ts` and replaced all three inline occurrences (two in `types.ts`, one in `ReviseoPanel.ts`) with the named alias.

PR #71, opened 2026-06-20
Save pr diff in seperate file instead of json

Sejdii:


src/core/review-generation/review/FetchReviewDataPhase.ts
```diff
@@ -12,21 +12,21 @@ export class FetchReviewDataPhase implements PromptComponent {
...
+Fetch the PR details and write them to the review data files:
 1. Fetch PR details via \`pull_request_read\` with method "get" for PR \`${this.prUrl}\` (to obtain title and description)
-2. Fetch the diff via \`pull_request_read\` with method "get_diff"
+2. Fetch the diff via \`pull_request_read\` with method "get_diff" and write it to \`${diffFilePath}\`
 3. Write \`${reviewDataPath}\` as a JSON file with this exact structure:
    {
      "title": "<PR title>",
      "description": "<PR description>",
-     "diff": "<full diff>"
+     "diff": "${diffFilePath}"
```
Sejdii: **[AI Generated by Clean Code Expert + Prompt Engineer + Skeptical Architect personas]**
**`diff` JSON field now stores a file path — rename or remove it entirely**

The resulting JSON schema is:
```json
{ "title": "...", "description": "...", "diff": ".ai/reviseo/1/diff.patch" }
```

The `diff` field has changed type from diff content to a filesystem path. This creates two concrete problems:

**1. Model misinterpretation (high-probability silent failure).** An AI model reading `review_data.json` will interpret `"diff": ".ai/reviseo/1/diff.patch"` as the actual diff content being the string `.ai/reviseo/1/diff.patch`. The name `diff` is semantically content-bearing — the model is likely to use the path string as the diff rather than reading the file it points to.

**2. A pointer where no pointer is needed.** Any consuming phase already knows the PR number and the base directory, so it can construct `.ai/reviseo/{prNumber}/diff.patch` directly. A pointer stored in a document where the reader already has enough information to derive the target is dead weight, and it creates a semantic contract (`diff` contains diff text) that the schema now violates.

**Recommended options, in order of preference:**

- **Rename the key to make the indirection explicit** if a pointer is genuinely useful: `"diffFilePath"` or `"diff_path"` removes all ambiguity for both human readers and AI agents.

The current approach is the worst of both worlds: two files to write, two files to clean up, and a type-confused JSON schema.

src/core/review-generation/review/FetchReviewDataPhase.ts
```diff
@@ -12,21 +12,21 @@ export class FetchReviewDataPhase implements PromptComponent {
 
     getText(): string {
         const reviewDataPath = `.ai/reviseo/${this.prNumber}/review_data.json`;
+        const diffFilePath = `.ai/reviseo/${this.prNumber}/diff.patch`;
```
Sejdii: **[AI Generated by Clean Code Expert persona]**
**Path construction duplicated across phase classes instead of using `REVISEO_BASE_DIR`**

`GithubReviewPromptFactory` already has a `REVISEO_BASE_DIR` constant and constructs `diffFilePath` using it. But both `FetchReviewDataPhase` and `CleanupPhase` independently reconstruct paths by hardcoding the `.ai/reviseo/` prefix inline:

```ts
const reviewDataPath = `.ai/reviseo/${this.prNumber}/review_data.json`;
const diffFilePath   = `.ai/reviseo/${this.prNumber}/diff.patch`;
```

This creates two sources of truth for the base directory. If the directory structure changes, the phase classes will not benefit from any update to `REVISEO_BASE_DIR`. The duplication is especially visible now that both `FetchReviewDataPhase` and `CleanupPhase` construct the same two paths using the same hardcoded prefix.

A cleaner design would pass the fully resolved file paths into the phase constructors as strings — the factory already builds them. The phases would then be free of any knowledge about the directory layout, which aligns with their actual responsibility: generating prompt text.

src/core/review-generation/review/CleanupPhase.ts
```diff
@@ -9,11 +9,12 @@ export class CleanupPhase implements PromptComponent {
...
     getText(): string {
         const reviewDataPath = `.ai/reviseo/${this.prNumber}/review_data.json`;
+        const diffFilePath = `.ai/reviseo/${this.prNumber}/diff.patch`;
         const condition = this.pendingReview
             ? 'once the pending review has been created and populated successfully'
             : 'once the review has been submitted successfully';
         return `## Phase ${this.phaseNumber}: Cleanup
 
-Delete \`${reviewDataPath}\` ${condition}.`;
+Delete \`${reviewDataPath}\` and \`${diffFilePath}\` ${condition}.`;
```
Sejdii: **[AI Generated by Skeptical Architect persona]**
**Two-file cleanup is non-atomic, and the two-file design adds complexity with no clear gain**

**Atomicity hazard.** The cleanup instruction now deletes `review_data.json` AND `diff.patch`. If the AI agent successfully deletes the first file but fails before deleting the second (network interruption, tool error, context exhaustion), the workspace is left with an orphaned `diff.patch` with no associated metadata. With a single-file approach the cleanup is one atomic operation: either the file is gone or it isn't.

**Design complexity.** The split introduces an extra file write, an extra file read, a semantic mutation of the `diff` JSON field (see separate comment), and a two-step cleanup — all to avoid embedding a string in JSON. The JSON's only remaining unique value over the patch file is `title` and `description`, which are short strings.

The industry-standard resolutions are simpler:
- **Drop `diff` from the JSON entirely.** Consuming phases construct the patch path by convention from the PR number they already hold. `review_data.json` becomes a pure metadata file; `diff.patch` is the diff. Cleanup deletes both — same two operations, but without the pointer anti-pattern.
- **Single structured file.** If the diff size is not a demonstrated problem at this scale (one PR at a time, bounded by the GitHub API response), keeping everything in one file eliminates the split entirely.

The only valid driver for this two-file split would be evidence that the JSON file was hitting token or file-size limits that caused real failures. There is no such evidence in this diff.

src/test/core/__fixtures__/multi-agent-two-personas.txt
```diff
@@ -6,25 +6,24 @@ You do not perform code review yourself — you delegate that work to the approp
...
-4. Confirm the file has been written successfully before reporting this phase complete.
+4. Confirm both files have been written successfully before reporting this phase complete.
 
 If any step fails, report the error and stop.
 
 ## Phase 2: Provide Multipersona Review
 
 For each persona step below, launch a reviewer subagent with:
-- The file path `.ai/reviseo/1/review_data.json` — the reviewer subagent must read this file to obtain the diff
+- The file path `.ai/reviseo/1/diff.patch` — the reviewer subagent must read this file to obtain the diff
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**Multi-agent subagents lose access to PR title and description**

Before this change, subagents received `review_data.json` which contained the PR title, description, and diff together. Now they are handed only `diff.patch`. Subagents have no shared context window with the orchestrator, so they cannot know the PR title or description — context that matters for checking whether the code change matches its stated intent.

The Phase 2 instruction currently reads:

```
- The file path `.ai/reviseo/1/diff.patch` — the reviewer subagent must read this file to obtain the diff
```

To restore title/description context, also pass the metadata file:

```
- The diff file `.ai/reviseo/1/diff.patch` — the reviewer subagent must read this to obtain the code changes
- The metadata file `.ai/reviseo/1/review_data.json` — the reviewer subagent must read this to obtain the PR title and description for context
```

Or, have the orchestrator read `review_data.json` at the start of Phase 2 and embed the title and description directly into each subagent's prompt string, which avoids an extra file read per subagent.

src/core/review-generation/review/FetchReviewDataPhase.ts
```diff
@@ -12,21 +12,21 @@ export class FetchReviewDataPhase implements PromptComponent {
...
+2. Fetch the diff via \`pull_request_read\` with method "get_diff" and write it to \`${diffFilePath}\`
 3. Write \`${reviewDataPath}\` as a JSON file with this exact structure:
    {
      "title": "<PR title>",
      "description": "<PR description>",
-     "diff": "<full diff>"
+     "diff": "${diffFilePath}"
    }
-4. Confirm the file has been written successfully before reporting this phase complete.
+4. Confirm both files have been written successfully before reporting this phase complete.
```
Sejdii: **[AI Generated by Prompt Engineer persona]**
**Step 4 confirmation is underspecified — "both files" is vague at the point of verification**

Step 4 says "Confirm both files have been written successfully" but at that point in the numbered list the two file paths appeared several steps earlier. Under any significant output length, the model may conflate which two files are meant or perform a shallow acknowledgement rather than an actual read-back check.

Make the verification step explicit about what to check:

```
4. Verify both files exist and are non-empty:
   - Read `.ai/reviseo/{prNumber}/diff.patch` and confirm it contains the raw diff
   - Read `.ai/reviseo/{prNumber}/review_data.json` and confirm it is valid JSON with non-empty "title", "description", and "diff_path" fields
   Only then report this phase complete.
```

This closes the gap between "I wrote the file" and "the file was written correctly", which matters because subsequent phases depend on both files existing with valid content.

src/core/review-generation/review/FetchReviewDataPhase.ts
Sejdii: Fixed — dropped the `diff` field from `review_data.json` entirely. The JSON is now pure metadata (`title` + `description` only), eliminating both the semantic confusion and the pointer anti-pattern. Consuming phases derive the patch path by convention from paths passed in at construction time.

src/core/review-generation/review/FetchReviewDataPhase.ts
Sejdii: Fixed — `FetchReviewDataPhase` and `CleanupPhase` constructors now accept fully-resolved `reviewDataPath` and `diffFilePath` strings. `GithubReviewPromptFactory` constructs both paths using `REVISEO_BASE_DIR` and passes them in, making it the single source of truth for directory layout.

src/core/review-generation/review/CleanupPhase.ts
Sejdii: Partially addressed — the two-file split is retained (it's the purpose of this PR), but the `diff`/`diffFilePath` pointer has been removed from `review_data.json`, eliminating the type-confused JSON schema and the redundant pointer. The JSON is now pure metadata; consuming phases derive the patch path by convention. The atomicity concern is acknowledged — it's accepted as a best-effort cleanup in a controlled single-agent environment.

src/test/core/__fixtures__/multi-agent-two-personas.txt
Sejdii: Fixed — `MultiAgentProvideMultipersonaReviewPhase` now accepts a `reviewDataPath` parameter and instructs each subagent to read both `diff.patch` (for code changes) and `review_data.json` (for PR title and description context).

src/core/review-generation/review/FetchReviewDataPhase.ts
Sejdii: Fixed — Step 4 now explicitly names both files and specifies what to verify in each: raw diff content in `diff.patch`, and valid JSON with non-empty `title` and `description` in `review_data.json`, before reporting the phase complete.

PR #73, opened 2026-06-21
Improve code review comments

Sejdii:


src/core/review-generation/review/CommentFormatComponent.ts
```diff
@@ -1,14 +1,27 @@
...
+3. What to do: exactly one recommendation. Show the proposed fix in code — never copy code from the PR, only the solution.
+   If a genuine alternative exists: one line starting with \`Alternative:\`.
+
+Total length: max 120 words. If more context is needed, use a collapsible block:
+<details>
+<summary>Full reasoning</summary>
+...
+</details>
+
+Deduplication: if two issues share the same root cause, write one comment. Comments may reference each other (e.g., "See also the comment on \`foo.ts:42\`") but must never duplicate content.`;
```
Sejdii: 🔴 [blocking] ❓ severity has no fix — "What to do" rule cannot be satisfied

The format mandates "What to do: exactly one recommendation. Show the proposed fix in code" for every comment. The ❓ `[question]` severity by definition has no fix — only a clarification request. The model will either fabricate a code snippet, skip the section, or produce malformed output.

Add a carve-out in the format spec:

```
What to do / What to ask:
- For ❓ comments: one precise question instead of a code fix.
- For all other severities: exactly one recommendation with a code snippet.
```

src/core/review-generation/review/PrepareCommentsPhase.ts
```diff
@@ -0,0 +1,25 @@
...
+import { CommentFormatComponent } from './CommentFormatComponent';
+
+export class PrepareCommentsPhase implements PromptComponent {
+    constructor(private readonly phaseNumber: number) {}
+
+    getText(): string {
+        const instructions =
+`## Phase ${this.phaseNumber}: Prepare Comments
+
+Transform the raw issues collected in Phase ${this.phaseNumber - 1} into formatted review comments:
```
Sejdii, line 11: 🟡 [suggestion] Inter-phase references via arithmetic break silently on reorder

Both `PrepareCommentsPhase` and `LocalWriteReviewPhase` reference their predecessor as `` Phase ${this.phaseNumber - 1} ``. If a phase is inserted or reordered, the generated prompt silently names the wrong phase — no compile-time error, no runtime warning, only a confused LLM.

Make the predecessor phase number an explicit constructor parameter:

```typescript
constructor(
  phaseNumber: number,
  private readonly issuePhaseNumber: number,
  // ...
) {}
// Usage: `Phase ${this.issuePhaseNumber}`
```

See also the same pattern in `LocalWriteReviewPhase.ts`.

src/core/review-generation/review/CommentFormatComponent.ts
```diff
@@ -1,14 +1,27 @@
 import type { PromptComponent } from './PromptComponent';
 
 const COMMENT_FORMAT =
-`### Comment format (MANDATORY)
+`### Comment format
```
Sejdii, line 4: 🟡 [suggestion] Dropping "(MANDATORY)" weakens format contract at worst moment

The heading was changed from `### Comment format (MANDATORY)` to `### Comment format` exactly when the format became more complex: 4 severity tiers, 3-part structure, 120-word cap, and collapsible blocks. Without the enforcement signal, a model may treat the detailed spec as advisory guidance rather than a strict output contract.

Restore the signal — either via the heading suffix or an opening sentence:

```markdown
### Comment format

Every comment **must** follow this structure exactly. Deviating from it is not permitted.
```

src/core/review-generation/review/CommentFormatComponent.ts
Sejdii: 🟡 [suggestion] Deduplication rule has two authoritative sources

The `Deduplication:` paragraph in `CommentFormatComponent` restates the same rule and cross-reference example already present in `PrepareCommentsPhase` step 2. Two copies of the same constraint will diverge as either component evolves independently.

Remove the `Deduplication:` section from `CommentFormatComponent` — the phase instructions own the grouping logic. `CommentFormatComponent` should describe structure only, not workflow.

Alternative: keep it only in `CommentFormatComponent` and remove it from the phase step.

src/core/review-generation/review/PrepareCommentsPhase.ts
```diff
@@ -0,0 +1,25 @@
...
+
+1. Perform root cause analysis: group issues that share the same underlying root cause.
+2. For each group (or standalone issue), write exactly one comment conforming to the format below.
+   - Comments may reference each other (e.g., "See also the comment on \`foo.ts:42\`") but must never duplicate content.
```
Sejdii: 🟡 [suggestion] File:line cross-references are stale before comments are submitted

The instruction tells the model to cross-reference using `` foo.ts:42 `` coordinates. At the PrepareComments stage no comments have been posted yet, so GitHub line anchors are not confirmed. Comments with `null` lines or adjusted positions will produce broken references.

Use description-based references instead:

```
Comments may reference each other (e.g., "See also the comment about
deduplication rule ownership") but must never duplicate content.
```

src/core/review-generation/review/SingleAgentFindIssuesPhase.ts
```diff
@@ -15,12 +15,12 @@ export class SingleAgentProvideMultipersonaReviewPhase extends ProvideMultiperso
...
-        return `## Phase ${this.phaseNumber}: Provide Multipersona Review
+        return `## Phase ${this.phaseNumber}: Find Issues
 
 Read \`${this.dataFilePath}\` to obtain the diff.${metadataLine}
-For each Step below, adopt that persona and perform your review according to its checklist.
-Generate review comments but do NOT submit them — submission happens in a later phase.
-Retain comments as an in-memory list; each item: persona name, file path, line number, title, body.
+For each Step below, adopt that persona and identify issues according to its checklist.
+Do NOT write formatted comments — comment preparation happens in a later phase.
+Retain issues as an in-memory list; each item: persona name, file path, line number, description.
```
Sejdii: 🔵 [nit] Single-agent description guidance less detailed than multi-agent

The single-agent retention instruction omits the richness qualifier that the multi-agent JSON schema carries: "enough context for the comment-preparation phase to write a quality comment without re-reading the diff." Without this, descriptions may be too thin for PrepareComments to produce quality output.

Append the same qualifier:

```
...description must capture the issue and its impact in enough detail
that PrepareComments can write a full comment without re-reading the diff.
```

src/core/review-generation/review/LocalWriteReviewPhase.ts
```diff
@@ -25,14 +21,17 @@ Base branch: \`${this.baseBranch}\`
 
 ## Findings
 
-### [Persona Name]
-**path/to/file.ts:line** — Finding title
-Finding body text.
+**path/to/file.ts:line**
+🔴 [blocking] concise description
+
```
Sejdii, line 33: 🔵 [nit] Output template implies all comments are blocking

The file template uses `🔴 [blocking] concise description` as the only severity placeholder. With four severity levels defined, the model may infer blocking is the expected default for all comments.

Show at least two different severity levels in the example:

```markdown
🔴 [blocking] description of a blocker
🟡 [suggestion] description of a recommendation
```

src/core/review-generation/review/CommentFormatComponent.ts
Sejdii: Fixed — added a `What to do / What to ask:` carve-out in the format spec: ❓ comments now call for one precise question instead of a code snippet, while all other severities keep the existing code-fix requirement.

src/core/review-generation/review/PrepareCommentsPhase.ts
Sejdii: Will not be applied — making `issuePhaseNumber` an explicit constructor parameter just moves the `n-1` arithmetic to the call site (`new PrepareCommentsPhase(n, n-1)`), providing no resilience against reordering. The phase ordering is already explicit in the factory arrays, making the relationship visible in context.

src/core/review-generation/review/CommentFormatComponent.ts
Sejdii: Fixed — added an opening enforcement sentence: "Every comment **must** follow this structure exactly. Deviating from it is not permitted."

src/core/review-generation/review/CommentFormatComponent.ts
Sejdii: Fixed — removed the `Deduplication:` section from `CommentFormatComponent`; the grouping rule now lives exclusively in `PrepareCommentsPhase` step 1–2 where the workflow logic belongs.

src/core/review-generation/review/PrepareCommentsPhase.ts
Sejdii: Fixed — switched to description-based cross-references: `"See also the comment about X"` instead of `"See also the comment on \`foo.ts:42\`"`.

src/core/review-generation/review/SingleAgentFindIssuesPhase.ts
Sejdii: Fixed — appended the qualifier: `description — the description must capture the issue and its impact in enough detail that PrepareComments can write a full comment without re-reading the diff.`

src/core/review-generation/review/LocalWriteReviewPhase.ts
Sejdii: Fixed — the file template now shows two entries with different severity levels (`🔴 [blocking]` and `🟡 [suggestion]`) so the model sees both as valid outputs.

PR #78, opened 2026-06-24
Refactor test assertions

Sejdii:


src/test/core/review-generation/ReviewPromptAssert.ts
```diff
@@ -0,0 +1,80 @@
...
+        const end = nextMatch ? nextMatch.index! : this._text.length;
+        return new PhaseAssert(this._text.slice(start, end).trim(), n);
+    }
+}
+
+class PhaseAssert {
+    constructor(private readonly _text: string, private readonly _n: number) {}
+
+    step(n: number): StepAssert {
+        const allMatches = [...this._text.matchAll(/^### Step (\d+):/gm)];
```
Sejdii: 🟡 [suggestion] Section-slicing logic is duplicated and silently fails on non-sequential phases

`phase()` and `step()` repeat the same regex-match-then-boundary-slice algorithm verbatim. Both sites share a silent failure mode: if phases are emitted out of numeric order, the wrong slice is extracted and assertions pass against incorrect content. A shared helper fixes both:
```ts
function extractSection(text: string, pattern: RegExp, n: number): string {
  const matches = [...text.matchAll(pattern)];
  const idx = matches.findIndex(m => Number(m[1]) === n);
  if (idx === -1) throw new Error(`Section ${n} not found`);
  const end = matches[idx + 1]?.index ?? text.length;
  return text.slice(matches[idx].index, end).trim();
}
```

src/test/core/review-generation/ReviewPromptAssert.ts
```diff
@@ -0,0 +1,80 @@
...
+        const match = allMatches.find(m => parseInt(m[1]) === n);
+        assert.ok(match, `Step ${n} not found in Phase ${this._n}`);
+        const start = match.index!;
+        const nextMatch = allMatches.find(m => parseInt(m[1]) > n);
+        const end = nextMatch ? nextMatch.index! : this._text.length;
+        return new StepAssert(this._text.slice(start, end).trim(), n, this._n);
+    }
+
+    hasTitle(title: string): this {
```
Sejdii: 🟡 [suggestion] Remove unused `hasTitle()` — YAGNI, no test exercises it

No test in this PR or the existing suite calls `hasTitle()`. Dead API on a freshly introduced helper signals speculative design and leaves future readers wondering what behaviour it guards. Remove it; re-add when a concrete test needs it.
```ts
// Remove:
hasTitle(title: string): this { ... }
```

src/test/core/review-generation/ReviewPromptAssert.ts
```diff
@@ -0,0 +1,80 @@
...
+        const header = `## Phase ${this._n}: ${title}`;
+        assert.ok(
+            this._text.startsWith(header),
+            `Phase ${this._n} expected title ${JSON.stringify(title)}, got: ${JSON.stringify(this._text.split('\n')[0])}`,
+        );
+        return this;
+    }
+
+    equalsText(expected: string): this {
+        assert.strictEqual(this._text, expected.trim(), `Phase ${this._n} text does not match expected`);
```
Sejdii: 🟡 [suggestion] `equalsText` silently trims expected string — rename or remove the trim

The method name implies exact equality, but the hidden `.trim()` means leading/trailing whitespace differences are invisible. If the prompt gains or loses a leading newline, the test still passes. Either rename to `equalsTextTrimmed`, or remove the trim and let each call site be explicit:
```ts
// Caller controls normalisation — intent is visible:
phase(4).equalsText(expected.trim());
```

src/test/core/review-generation/ReviewPromptAssert.ts
```diff
@@ -0,0 +1,80 @@
...
+        return this;
+    }
+
+    contains(text: string): this {
```
Sejdii, line 39: 🔵 [nit] `contains()` body is duplicated verbatim in `PhaseAssert` and `StepAssert`

Both call `assert.ok(this._text.includes(text), ...)` differing only in the error prefix. If whitespace-normalisation is ever needed, both sites require the same edit. Consider a shared private helper or a common base.

src/test/core/review-generation/ReviewPromptAssert.ts
```diff
@@ -0,0 +1,80 @@
+import * as assert from 'assert';
+
+export function assertPrompt(prompt: string): ReviewPromptAssert {
+    return new ReviewPromptAssert(prompt);
+}
+
+class ReviewPromptAssert {
+    constructor(private readonly _text: string) {}
```
Sejdii: 🔵 [nit] Underscore-prefixed private fields are redundant with TypeScript `private`

`_text`, `_n`, `_phaseN` use a pre-TypeScript convention for signalling privacy by convention. The `private` modifier already enforces access control at compile time. Rename to `text`, `n`, `phaseN` to reduce visual noise.

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: Fixed — extracted a shared `extractSection` helper that uses `findIndex` + `matches[idx + 1]` to get the next boundary, which correctly handles non-sequential phase/step numbers.

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: Fixed — `hasTitle()` removed; no test in the suite uses it and it aligns with the project's YAGNI principle.

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: Fixed — removed the hidden `.trim()` from `equalsText`; the call site now explicitly passes `.trim()`, making the whitespace normalisation visible in context.

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: Fixed — extracted a private `assertContains` module-level function shared by both `PhaseAssert` and `StepAssert`, so any future whitespace-normalisation change lives in one place.

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: Fixed — removed the `_` prefix from all private fields; TypeScript's `private` modifier is sufficient.

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: 🔵 [nit] The error message `Section ${n} not found` is still too generic — it doesn't distinguish whether a phase or a step was missing. Suggest accepting a label parameter so failures are self-diagnosing:
```ts
function extractSection(text: string, pattern: RegExp, n: number, label: string): string {
    const matches = [...text.matchAll(pattern)];
    const idx = matches.findIndex(m => Number(m[1]) === n);
    assert.ok(idx !== -1, `${label} ${n} not found`);
    ...
}
// callsite: extractSection(this.text, /^## Phase (\d+):/gm, n, 'Phase')
// callsite: extractSection(this.text, /^### Step (\d+):/gm, n, 'Step')
```

src/test/core/review-generation/ReviewPromptAssert.ts
```diff
@@ -0,0 +1,68 @@
...
+        );
+        return this;
+    }
+
+    contains(text: string): this {
+        assertContains(this.text, text, `Step ${this.n} in Phase ${this.phaseN}`);
+        return this;
+    }
+
+    and(): this {
```
Sejdii: 🔵 [nit] `and()` no-op implies behavior it doesn't have

A method named `and()` that silently returns `this` without any validation suggests it enforces a boundary or transitions state. Readers stop and trace through it expecting something to happen. It adds visual noise without semantic value, and will confuse contributors who wonder what invariant it guards.

Remove it and chain methods directly:
```ts
step(1).hasName('Security Auditor').contains('focus area');
```

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: Fixed — added a `label` parameter to `extractSection` and updated both call sites to pass `'Phase'` and `'Step'` respectively, so failures now report "Phase 3 not found" or "Step 2 not found".

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: Fixed — removed `and()` from `StepAssert` and updated all three call sites to chain methods directly.

PR #79, opened 2026-06-27
Prefer to add comment to specific line

Sejdii:


src/core/review-generation/review/SingleAgentFindIssuesPhase.ts
```diff
@@ -20,7 +20,7 @@ export class SingleAgentFindIssuesPhase extends FindIssuesPhase {
 Read \`${this.dataFilePath}\` to obtain the diff.${metadataLine}
 For each Step below, adopt that persona and identify issues according to its checklist.
 Do NOT write formatted comments — comment preparation happens in a later phase.
-Retain issues as an in-memory list; each item: persona name, file path, line number, description — the description must capture the issue and its impact in enough detail that PrepareComments can write a full comment without re-reading the diff.
+Retain issues as an in-memory list; each item: persona name, file path, line number (an integer from the diff — prefer a specific line even for broad issues; use null only when no single changed line is representative of the problem), description — the description must capture the issue and its impact in enough detail that PrepareComments can write a full comment without re-reading the diff.
```
Sejdii: 🟡 [suggestion] extract duplicated line-number guidance to a shared constant

The line-number rule is copy-pasted verbatim into both `SingleAgentFindIssuesPhase` and `MultiAgentFindIssuesPhase` with structurally different forms (a 25-word prose parenthetical vs a JSON field comment). A future policy change requires finding and updating both, and the structural difference makes a grep unreliable. Embedding the constraint mid-sentence also makes it easy for the model to treat it as an aside rather than a binding rule.

Extract to a shared constant and interpolate in both phases; in the single-agent form break it out as its own field note rather than a parenthetical:

```ts
const LINE_NUMBER_GUIDANCE = `prefer a specific line even for broad issues; use null only when no single changed line is representative of the problem`;
```

src/core/review-generation/review/MultiAgentFindIssuesPhase.ts
```diff
@@ -29,7 +29,7 @@ Every element must conform to:
 {
   "persona": "<persona name>",
   "file": "<path/to/file.ext>",
-  "line": <integer line number, or null for file-level issues>,
+  "line": <integer line number from the diff — prefer a specific line even for broad issues; use null only when no single changed line is representative of the problem>,
```
Sejdii: 🔴 [blocking] "line number from the diff" conflates diff position with new-file line

Unified diffs expose two coordinate systems: the hunk offset (sequential position within the hunk block) and the new-file line number (the right-hand value in the `@@` header). GitHub's inline comment API expects the new-file line number. A model that outputs a hunk offset instead will place every comment on the wrong line.

Rewrite as: `the new-file line number (right-hand number in the @@ header, 1-indexed) of the changed line most relevant to this issue`.

Note: the same fix is needed in `SingleAgentFindIssuesPhase.ts`.

src/core/review-generation/review/SubmitReviewPhase.ts
```diff
@@ -18,7 +18,7 @@ export class SubmitReviewPhase implements PromptComponent {
 ${header}
 
 1. Create a new pending review. Do not add a review body — leave it empty.
-2. Add each prepared comment to the pending review using available tools.
+2. Add each prepared comment to the pending review: use the inline comment tool (specifying file path and line number) when a line number is set; use the file-level comment tool only when the line is null.
```
Sejdii: 🟡 [suggestion] step 2 conflates two cases and names tools by abstract role

Step 2 packs two branching cases into one run-on sentence and refers to tools as "inline comment tool" / "file-level comment tool" without tying them to actual parameter signatures. When the model's available tools are named `add_pull_request_review_comment` and `create_review_comment`, it must guess which accepts a `line` parameter — and may guess wrong.

Split into sub-items and identify tools by their distinguishing parameter:

```
2a. line is set → use the tool that accepts `path` + `line` parameters.
2b. line is null → use the tool that accepts only `path` (no line).
```

src/core/review-generation/review/PrepareCommentsPhase.ts
```diff
@@ -28,7 +28,8 @@ ${filteringBlock}Transform the raw issues collected in Phase ${this.phaseNumber
 1. Perform root cause analysis: group issues that share the same underlying root cause.
 2. For each group (or standalone issue), write exactly one comment conforming to the format below.
    - Comments may reference each other (e.g., "See also the comment about X") but must never duplicate content.
-3. Retain the final comment list for submission in the next phase.
+3. Resolve line targets: for each comment whose source issue had no line number, identify the changed line in the diff most directly related to the comment's concern and attach the comment there. Keep a comment file-level only when the issue genuinely applies to the whole file and no single changed line is meaningfully more relevant than others.
```
Sejdii: 🟡 [suggestion] step 3 line resolution is underspecified and lacks a tie-breaking rule

Step 3 asks PrepareComments to re-examine the diff for null-line issues, but in multi-agent mode this phase runs in a separate agent context that may not have the diff path in scope. There is also no tie-breaking rule for when multiple changed lines are equally plausible, making placement non-deterministic across runs.

PrepareCommentPhase should belong on data provided by finding issue phase

src/core/review-generation/review/PrepareCommentsPhase.ts
```diff
@@ -28,7 +28,8 @@ ${filteringBlock}Transform the raw issues collected in Phase ${this.phaseNumber
...
+4. Retain the final comment list (with file path and resolved line number) for submission in the next phase.
```
Sejdii: 🟡 [suggestion] step 4 omits the carry-over schema, risking structural drift between runs

Step 4 says "retain the final comment list" without specifying its structure. FindIssues produced a JSON array with a defined schema; PrepareComments transforms those and must pass a predictable structure to SubmitReview. Without an explicit schema, the model invents a format each run, making SubmitReview's input non-deterministic.

Specify the format explicitly:

```
Retain as: [{ file: string, line: integer | null, body: string }]
```

src/core/review-generation/review/SingleAgentFindIssuesPhase.ts
Sejdii: Applied — extracted `LINE_NUMBER_GUIDANCE` as an exported constant in `FindIssuesPhase.ts` and imported it in both subclasses. Also restructured the single-agent form from a run-on parenthetical into bullet-point field notes so the guidance is a first-class item rather than an aside.

src/core/review-generation/review/MultiAgentFindIssuesPhase.ts
Sejdii: Applied — replaced "integer line number from the diff" with "new-file line number (right-hand number in the @@ header, 1-indexed)" in both `SingleAgentFindIssuesPhase` and `MultiAgentFindIssuesPhase` via the shared `LINE_NUMBER_GUIDANCE` constant.

src/core/review-generation/review/SubmitReviewPhase.ts
Sejdii: Applied — split step 2 into sub-items 2a/2b, identifying each branch by its distinguishing parameter (`path` + `line` vs `path` only) rather than by abstract role name.

src/core/review-generation/review/PrepareCommentsPhase.ts
Sejdii: Partially applied — added the tie-breaking rule to step 3 ("if multiple changed lines are equally plausible, use the first one"). The multi-agent context issue (PrepareComments not having the diff in scope) is deferred: fixing it properly means either passing the diff path into `PrepareCommentsPhase` or having `FindIssues` always resolve line numbers itself, both of which are broader changes outside this PR's scope.

src/core/review-generation/review/PrepareCommentsPhase.ts
Sejdii: Applied — step 4 now specifies the carry-over format explicitly: `[{ file: string, line: integer | null, body: string }]`.

PR #80, opened 2026-06-27
Add option to skip cleanup phase to keep files

Sejdii:


src/test/core/review-generation/ReviewPromptAssert.ts
```diff
@@ -22,6 +22,12 @@ class ReviewPromptAssert {
     phase(n: number): PhaseAssert {
         return new PhaseAssert(extractSection(this.text, /^## Phase (\d+):/gm, n, 'Phase'), n);
     }
+
+    doesntHavePhase(name: string): this {
```
Sejdii: 🔵 [nit] `doesntHavePhase` breaks DSL convention and its regex is unsafe

The DSL uses clean verb phrases (`hasName`, `contains`); `doesntHavePhase` is an informal contraction that reads out of place alongside them. Additionally, `name` is interpolated into a `RegExp` without escaping — a phase name with metacharacters would silently mismatch.

Rename to `hasNoPhase` and escape the input:
```ts
hasNoPhase(name: string): this {
    const escaped = name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
    const pattern = new RegExp(`^## Phase \\d+: ${escaped}$`, 'm');
    assert.ok(!pattern.test(this.text), `Expected no phase named "${name}" but found one`);
    return this;
}
```

src/core/review-generation/LocalReviewPromptFactory.ts
```diff
@@ -33,7 +33,7 @@ export class LocalReviewPromptFactory implements ReviewPromptFactory {
                 : new SingleAgentFindIssuesPhase(n, stepsComponent, fullPath),
             (n) => new PrepareCommentsPhase(n),
             (n) => new LocalWriteReviewPhase(n, config.timestamp, config.baseBranch),
-            (n) => new LocalCleanupPhase(n, fullPath),
+            ...(config.skipCleanup ? [] : [(n: number) => new LocalCleanupPhase(n, fullPath)]),
```
Sejdii: 🟡 [suggestion] Inline lambda in conditional spread requires an annotation absent from all peer entries

TypeScript loses inference inside the conditional array, forcing an explicit `(n: number)` annotation that none of the sibling entries carry. `GithubReviewPromptFactory` avoids this by pre-declaring `cleanupPhaseFactory` as a named variable before the array. Apply the same pattern here:
```ts
const cleanupPhaseFactory = (n: number) => new LocalCleanupPhase(n, fullPath);
// then in phaseFactories:
...(config.skipCleanup ? [] : [cleanupPhaseFactory]),
```

src/extension/webview/ReviseoPanel.ts
```diff
@@ -105,23 +106,24 @@ export class ReviseoPanel {
...
                     context: message.personaContext ?? {},
+                    skipCleanup: message.skipCleanup,
                 };
                 this._panel.webview.postMessage({ type: 'promptGenerated', text: this._reviewGen.build(config) });
                 break;
             }
             case 'getInitialState': {
-                const github = this._context.globalState.get<{ multiAgent: boolean; pendingReview: boolean; skipCommentedIssues: boolean }>(REVIEW_SETTINGS_KEY, { multiAgent: false, pendingReview: false, skipCommentedIssues: false });
-                const local = this._context.workspaceState.get<{ multiAgent: boolean; baseBranch: string }>(LOCAL_REVIEW_SETTINGS_KEY, { multiAgent: false, baseBranch: DEFAULT_BASE_BRANCH });
+                const github = this._context.globalState.get<{ multiAgent: boolean; pendingReview: boolean; skipCommentedIssues: boolean; skipCleanup: boolean }>(REVIEW_SETTINGS_KEY, { multiAgent: false, pendingReview: false, skipCommentedIssues: false, skipCleanup: false });
```
Sejdii: 🟡 [suggestion] Settings shape has no named type — each new field requires touching 6 anonymous sites

This PR added `skipCleanup` to six anonymous inline types: two `get<{...}>` generics, two default literals, two `update()` calls. Without a named type these sites drift silently — and the compiler won't catch a default value that's out of sync with the generic parameter.

Define them once:
```ts
type GithubReviewSettings = { multiAgent: boolean; pendingReview: boolean; skipCommentedIssues: boolean; skipCleanup: boolean };
type LocalReviewSettings  = { multiAgent: boolean; baseBranch: string; skipCleanup: boolean };
```
Alternative: move both to `webview/types.ts` alongside the other message types.

src/extension/webview/types.ts
```diff
@@ -8,12 +8,12 @@ export type WebviewMessage =
     | { type: 'savePersona'; persona: Persona }
     | { type: 'deletePersona'; id: string }
     | { type: 'generatePrompt'; prUrl: string; personaIds: string[]; promptOptions: PromptOptions; personaContext?: Record<string, Record<string, string>> }
```
Sejdii, line 10: 🟡 [suggestion] `generateLocalPrompt` spreads options as flat fields while `generatePrompt` uses `PromptOptions`

`generatePrompt` wraps options in `promptOptions: PromptOptions`; `generateLocalPrompt` passes `multiAgent` and `skipCleanup` as flat top-level fields. Each new shared option now requires parallel additions in both message variants and both handlers in `ReviseoPanel`. `PromptOptions` was introduced to be the single source of truth for these flags, but the local path bypasses it entirely.

Align the local path with a structured options bag:
```ts
| { type: 'generateLocalPrompt'; baseBranch: string; personaIds: string[]; promptOptions: LocalPromptOptions; personaContext?: Record<string, Record<string, string>> }
```

src/test/core/review-generation/ReviewPromptAssert.ts
Sejdii: Fixed — renamed to `hasNoPhase` to match the DSL's verb-phrase conventions, and added regex escaping on the input before building the pattern.

src/core/review-generation/LocalReviewPromptFactory.ts
Sejdii: Fixed — pre-declared `cleanupPhaseFactory` as a named variable before the array, matching the pattern used in `GithubReviewPromptFactory`.

src/extension/webview/types.ts
Sejdii: Will not be applied in this PR — the concern is valid, but `LocalPromptOptions` doesn't exist yet and aligning the two paths requires a new type plus changes across `types.ts`, `reviewScript.ts`, and `ReviseoPanel.ts`. That's a standalone refactor; bundling it here would widen the scope of what is intentionally a focused change. Happy to track it as a follow-up.

PR #82, opened 2026-06-27
Fetch PR data using script and gh cli

Sejdii:


src/extension/webview/ReviseoPanel.ts
```diff
@@ -141,6 +160,27 @@ export class ReviseoPanel {
...
+                const workspaceFolders = vscode.workspace.workspaceFolders;
+                if (!workspaceFolders || workspaceFolders.length === 0) {
+                    void vscode.window.showErrorMessage('Reviseo: No workspace folder is open. Please open a folder before fetching PR data.');
+                    break;
+                }
+                const prNumber = message.prUrl.match(/\/pull\/(\d+)/)?.[1] ?? '0';
+                const jsonFields = message.skipCommentedIssues ? 'title,body,reviewThreads' : 'title,body';
+                const commands = [
+                    `mkdir -p .ai/reviseo/${prNumber}`,
+                    `gh pr diff ${message.prUrl} > .ai/reviseo/${prNumber}/diff.patch`,
```
Sejdii, line 151: 🔴 [blocking] user-supplied prUrl interpolated verbatim into shell command

`message.prUrl` comes from an unvalidated text input and is placed directly into the string passed to `terminal.sendText`. A URL containing shell metacharacters — e.g. `https://github.com/org/repo/pull/1; rm -rf ~` — executes the injected command in the user's shell without any escaping or allowlist check.

Validate the URL with a strict pattern and reject anything that doesn't match:
```typescript
if (!/^https:\/\/github\.com\/[^/\s]+\/[^/\s]+\/pull\/\d+$/.test(message.prUrl)) {
    void vscode.window.showErrorMessage('Reviseo: Invalid PR URL');
    break;
}
```

src/extension/webview/ReviseoPanel.ts
```diff
@@ -141,6 +160,27 @@ export class ReviseoPanel {
...
                 break;
             }
+            case 'fetchReviewData': {
+                const workspaceFolders = vscode.workspace.workspaceFolders;
+                if (!workspaceFolders || workspaceFolders.length === 0) {
+                    void vscode.window.showErrorMessage('Reviseo: No workspace folder is open. Please open a folder before fetching PR data.');
+                    break;
+                }
+                const prNumber = message.prUrl.match(/\/pull\/(\d+)/)?.[1] ?? '0';
+                const jsonFields = message.skipCommentedIssues ? 'title,body,reviewThreads' : 'title,body';
```
Sejdii: 🔴 [blocking] `reviewThreads` is not a valid `gh pr view --json` field — command always fails when `skipCommentedIssues` is true

`gh pr view --json` accepts a fixed whitelist of ~46 PR fields; `reviewThreads` is not among them (confirmed by running `gh pr view --json` with no args). The command exits non-zero, and the `>` redirect creates an **empty** `review_data.json` (errors go to stderr). `fs.access` then succeeds on the empty file, `skipPrDataFetchPhase` is set to `true`, and the AI is told to parse an empty file while the fetch phase is skipped entirely.

Use: `gh api repos/{owner}/{repo}/pulls/{number}/comments` instead

src/core/review-generation/review/FetchReviewDataPhase.ts
```diff
@@ -11,11 +11,11 @@ export class FetchReviewDataPhase implements PromptComponent {
...
         const existingCommentsSteps = this.skipCommentedIssues
-            ? `4. Fetch existing unresolved inline review comments via \`pull_request_read\` with method "list_review_comments" for PR \`${this.prUrl}\`. For each comment retain only: id, path, line, body. Filter to unresolved comments only. Re-write \`${this.reviewDataPath}\` adding an "existingComments" array field to the existing JSON (preserve "title" and "description"). If there are no unresolved comments, write "existingComments": [].
-5. Verify \`${this.reviewDataPath}\` is valid JSON with "title", "description", and "existingComments" fields present.`
+            ? `4. Fetch existing unresolved inline review comments via \`pull_request_read\` with method "list_review_comments" for PR \`${this.prUrl}\`. Group each comment into a thread object: \`{ "isResolved": false, "path": "<file path>", "line": <line number>, "comments": [{ "id": "<comment id>", "body": "<comment body>" }] }\`. Filter to unresolved comments only. Re-write \`${this.reviewDataPath}\` adding a "reviewThreads" array field to the existing JSON (preserve "title" and "body"). If there are no unresolved comments, write "reviewThreads": [].
```
Sejdii: 🟡 [suggestion] `"isResolved": false` hardcoded in every thread makes `PrepareCommentsPhase`'s filter a permanent no-op

The template instructs the AI to construct every thread object with `"isResolved": false`. Since all threads always have that value by construction, the filter in `PrepareCommentsPhase` — "filter to threads where `isResolved` is false" — passes every thread unconditionally. Resolved comments are never filtered out.

There is also no data available to set this correctly: `list_review_comments` returns individual comment objects, not thread-level resolution state (that requires the GraphQL API). The AI has no signal to distinguish resolved from unresolved at write time.

Either switch to the GraphQL API to fetch threads with real resolution state, or remove the `isResolved` field and the filter entirely until that data source is available.

src/extension/webview/ReviseoPanel.ts
```diff
@@ -83,18 +85,35 @@ export class ReviseoPanel {
...
-                    url: message.prUrl,
-                    personas: selected,
-                    personaReviewExecutionMode: message.promptOptions.multiAgent ? PersonaReviewExecutionMode.MULTI_AGENT : PersonaReviewExecutionMode.SINGLE_AGENT,
-                    context: message.personaContext ?? {},
-                    pendingReview: message.promptOptions.pendingReview,
-                    skipCommentedIssues: message.promptOptions.skipCommentedIssues,
-                    skipCleanup: message.promptOptions.skipCleanup,
-                };
-                this._panel.webview.postMessage({ type: 'promptGenerated', text: this._reviewGen.build(config) });
+                void (async () => {
```
Sejdii: 🟡 [suggestion] `void` async IIFE swallows rejections — webview hangs with no error feedback if anything throws

`void (async () => { ... })()` discards the promise. Any unhandled rejection inside — from a future validation in `build()`, an unexpected `fs` behaviour, or any new code added to this block — is silently dropped. The webview never receives `promptGenerated` and appears frozen with no indication of what went wrong.

Extract to a named private method and handle the error explicitly:
```typescript
private async _handleGeneratePrompt(message: ...) { /* ... */ }

// in switch:
this._handleGeneratePrompt(message).catch(err => {
    console.error('reviseo: prompt generation failed', err);
    void vscode.window.showErrorMessage(`Reviseo: failed to generate prompt`);
});
```

src/extension/webview/ReviseoPanel.ts
```diff
@@ -83,18 +85,35 @@ export class ReviseoPanel {
...
+                    const selected = this._store.getAll().filter(p => message.personaIds.includes(p.id));
+                    const prNumber = message.prUrl.match(/\/pull\/(\d+)/)?.[1] ?? '0';
```
Sejdii: 🔵 [nit] PR number extraction regex appears three times across two layers with a silent `'0'` fallback

`message.prUrl.match(/\/pull\/(\d+)/)?.[1] ?? '0'` is duplicated here (line 90), at line 169, and in `GithubReviewPromptFactory.ts`. The `'0'` fallback silently proceeds to create `.ai/reviseo/0/` and run `gh` commands against PR 0 when the URL doesn't match — surfacing only an opaque CLI error.

Extract to a shared helper that returns `null` on mismatch, so both call sites can surface an explicit error:
```typescript
function extractPrNumber(url: string): string | null {
    return url.match(/\/pull\/(\d+)/)?.[1] ?? null;
}
```

src/extension/webview/ReviseoPanel.ts
Sejdii: Fixed — added a strict regex guard (`/^https:\/\/github\.com\/[^/\s]+\/[^/\s]+\/pull\/\d+$/`) at the top of the `fetchReviewData` case that rejects the URL with an error message before any shell command is constructed.

src/extension/webview/ReviseoPanel.ts
Sejdii: Fixed — removed `reviewThreads` from `gh pr view --json` (now always `title,body`). Since review threads can only be fetched correctly via MCP in `FetchReviewDataPhase` step 4, the `generatePrompt` handler now only sets `skipPrDataFetchPhase=true` when `skipCommentedIssues` is false — ensuring Phase 1 always runs when thread data is needed.

src/core/review-generation/review/FetchReviewDataPhase.ts
Sejdii: Fixed — removed `"isResolved": false` from the thread object template in `FetchReviewDataPhase` and removed the "filter to threads where `isResolved` is false" instruction from `PrepareCommentsPhase`. The REST API doesn't provide resolution state, so we no longer pretend it does. Fixtures updated to match.

src/extension/webview/ReviseoPanel.ts
Sejdii: Fixed — extracted the async logic into a private `_handleGeneratePrompt` method and wired it up with `.catch()` that logs the error and shows a `showErrorMessage` to the user.

src/extension/webview/ReviseoPanel.ts
Sejdii: Fixed — extracted `extractPrNumber(url: string): string | null` into `src/core/review-generation/PrUrlUtils.ts`, exported it from the module's `index.ts`, and replaced all three call sites. The `'0'` silent fallback is gone; all callers now handle the `null` case explicitly (early return in `GithubReviewPromptFactory`, error message + break in `fetchReviewData`, `null`-safe guard before the file-existence check in `_handleGeneratePrompt`).

